PR #3, opened 2019-08-08
feat(tooltip-signpost): added tooltip and signpost

raizen62:
Tooltip : 
psHKXm6OJX
Signpost:
61VoCuOOh0


libs/angular-components/src/signpost/src/signpost.component.html
```diff
@@ -0,0 +1,3 @@
+<button id="signpost" mat-icon-button>
+  <mat-icon>info</mat-icon>
```
rbordeanu: let's use the `info_outlined` icon, I think it looks better

libs/angular-components/src/signpost/src/signpost.component.html
```diff
@@ -0,0 +1,3 @@
+<button id="signpost" mat-icon-button>
```
rbordeanu: Not sure if it should be a button, or a more "discrete" element: i.e simple div with primary color on hover?

libs/angular-components/src/tooltip/src/tooltip.directive.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.loadTippy();
+  }
+
+  loadTippy() {
+    setTimeout(() => {
+      const el = this.el.nativeElement;
+      console.log(el);
+      const tippyOptions = {
+        placement: this.position || 'top',
+        content: this.content || 'Default tooltip'
```
rbordeanu: Should not have default content, throw an error if content is not provided.

libs/angular-components/src/signpost/src/signpost.component.ts
```diff
@@ -0,0 +1,39 @@
...
+  @Input() content;
+  @Input() position;
+  ngOnInit() {
+    this.loadTippy();
+  }
+
+  loadTippy() {
+    setTimeout(() => {
+      const tippyOptions =  {
+        content: this.content || 'Hello',
```
rbordeanu: Should not have default content, throw an error if content is not provided.

libs/angular-components/src/signpost/src/signpost.component.ts
```diff
@@ -0,0 +1,39 @@
...
+        placement: this.position || 'right',
```
rbordeanu: Should default to 'top'.

libs/angular-components/src/signpost/src/signpost.component.ts
rbordeanu: Also, make sure that content can have safe html, like clickable links.

themes/angular-theme/lib/tooltip/_tooltip-theme.scss
```diff
@@ -0,0 +1,27 @@
+
```
rbordeanu: Don't hardcode color values or anything :) 
Please see best practices for defining a component theme: https://github.com/fusionfabric/finastra-design-system/blob/master/themes/angular-theme/lib/button/_button-theme.scss#L2
(for example)
Define a mixin, use custom variables, extract known values from theme object etc.

libs/angular-components/src/signpost/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@ffdc/uxg-angular-components/tooltip",
+  "version": "0.0.1"
```
aifrim: Due to recent changes you will need to manually update version to `0.12.0`

themes/angular-theme/lib/core/theming/_all-theme.scss
```diff
@@ -11,6 +11,7 @@
 @import '../../menu/menu-theme';
 @import '../../sidenav/sidenav-theme';
 @import '../../toolbar/toolbar-theme';
+@import '../../tooltip/tooltip-theme';
```
aifrim: We will need to remove this. @ttalbot can give you more information.

themes/angular-theme/lib/tooltip/_tooltip-theme.scss
aifrim: This file needs to be removed from the theme and added to the component.

aifrim:
> My opinion is that both tooltip and signpost should be part of the theme and based on Angular-Material tooltip.

We should discuss this during the next FDS Sync

rbordeanu:
> My opinion is that both tooltip and signpost should be part of the theme and based on Angular-Material tooltip.

Well I disagree with that.

For once, the Material tooltip is very limiting.
As per spec: 
_Tooltips don’t display rich information including images and formatted text.
Tooltips are different than ALT-attributes, which are intended primarily for static images.
Tooltips don’t have directional arrows; instead, they rely on motion emanating from the source to convey direction._

We DO want to have rich content (HTML) inside the signpost popover, as well as the arrows indicating origin.

Tippy is a really small dependency, with many customisable options, and it looks great (subjective opinion but quite popular).

We'd like to have ready-made directives handling tooltips and signposts to be used directly by developers in a consistent way, without any extra work: like `uxgTooltip` and `uxgSignpost content="whatever"`.

aifrim:
Closing this PR because it is obsolete.

PR #4, opened 2019-08-12
bugfix(angular-theme): cleanup dependencies

bcldvd:
As per title.

### DIFF
--- a/themes/angular-theme/package.json
+++ b/themes/angular-theme/package.json
@@ -10,7 +10,6 @@
     "@angular/cdk": "^8.0.1",
     "@angular/common": "~8.0.2",
     "@angular/core": "~8.0.2",
-    "@angular/forms": "^8.0.2",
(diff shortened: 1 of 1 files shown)


themes/angular-theme/package.json
```diff
@@ -6,13 +6,6 @@
   "sass": "_all-theme.scss",
   "css": "_all-theme.css",
   "dependencies": {
```
aifrim: Angular asks us to have `peerDependencies` when creating reusable components. Should we do this as well here?

bcldvd: Either we create a custom schematics to add the dependency in the project, or we add a dependency and not peerDependency.
Knowing that and our bandwidth, I would be inclined to have dependencies for the time being, and why not change it in a later future if we create a schematics to `ng add` the theme/components

PR #5, opened 2019-08-12
feat(github): add CODEOWNERS file to enforce reviews by owners

ttalbot:
As documented in Github [here](https://help.github.com/en/articles/about-code-owners), we can request review from owners of some files / folders of a repo.
Just testing the feature at the moment before adding more owners.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,8 @@
+# Themes and style files should be reviewed by Studios' team.
+/themes/    @ttalbot
+*.scss      @ttalbot
+*.css       @ttalbot
+*.png       @ttalbot
+*.jpeg      @ttalbot
+*.jpg       @ttalbot
+*.svg       @ttalbot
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# Themes and style files should be reviewed by Studios' team.
```
aifrim: Good idea.

We need to discuss who are the codeowners of the other parts of the solution.

aifrim:
@bcldvd should we update the schematic to also add the components creator to the `CODEOWNERS` file?

.github/CODEOWNERS
ttalbot: Global codeowners added as well as repo management and per components owners.

.github/CODEOWNERS
```diff
@@ -0,0 +1,20 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence.
+@aifrim @bcldvd @ttalbot
```
aifrim: [Here](https://help.github.com/en/articles/about-code-owners#example-of-a-codeowners-file) it states that global codeowners have to be added as so:
```
* @aifrim @bcldvd @ttalbot 
```

ttalbot: right, thanks. Fixed

PR #14, opened 2019-08-29
fix(icon): add uxg icons sizes and colors in angular theme

ttalbot:
Screenshot 2019-08-29 at 10 32 39

### DIFF
--- a/themes/angular-theme/lib/icon/_icon-theme.scss
+++ b/themes/angular-theme/lib/icon/_icon-theme.scss
@@ -1,0 +1,7 @@
+@mixin uxg-icon-theme($theme) {
+    .mat-icon[color=gradient] {
+        background: mat-color($uxg-gradient, vertical);
+        background-clip: text;
+        -webkit-text-fill-color: transparent;
+    }
+}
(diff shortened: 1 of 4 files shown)


themes/angular-theme/lib/icon/_icon-theme.scss
```diff
@@ -0,0 +1,7 @@
+@mixin uxg-icon-theme($theme) {
+  .mat-icon[color=gradient] {
+    background: mat-color($uxg-gradient, vertical);
+    background-clip: text;
+    -webkit-text-fill-color: transparent;
```
aifrim: [-webkit-text-fill-color Browser Compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color#Browser_compatibility) has no support for IE.

Will the icon be displayed accordingly in IE 11?

rbordeanu:
Aha! So you added extra-large after all 😄

themes/angular-theme/lib/icon/_icon-theme.scss
ttalbot: Added media query targeting IE10+.
Apply accent color when gradient is not possible.

PR #15, opened 2019-08-29
feat(typography): update typography scales

ttalbot:
update typography scales to reflect changes made in the Design System.

### DIFF
--- a/themes/angular-theme/lib/avatar/_avatar-theme.scss
+++ b/themes/angular-theme/lib/avatar/_avatar-theme.scss
@@ -48,17 +48,17 @@
 
     &[dense] {
       font: {
-        family: mat-font-family($config, body-3);
-        size: mat-font-size($config, body-3);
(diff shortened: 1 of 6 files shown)


aifrim:
This is a general question about those variable names.

We have `display1-6`, `body1-6` and others like this.
Shouldn't we name them something more... proper? Something that applies within the current branding & design guidelines?

aifrim:
> > This is a general question about those variable names.
> > We have `$display1-6`, `$body1-6` and others like this.
> > Shouldn't we name them something more... proper? Something that applies within the current branding & design guidelines?
> 
> Those terms already carry lots of meaning in [material](https://material.io/design/typography/the-type-system.html#applying-the-type-scale).
> Could you give an example of what you would feel more proper?

You are right. I need to be more in tune with these.
Thank you for letting me know 😄 👍 ❤️

PR #16, opened 2019-08-30
bugfix(formatting): fix project formatting

aifrim:
Used `npm run format` to properly format all the code.
All present PRs will have to run this before committing.

**The actual command I used was**
```sh
nx format:write --base=b86e935dbd3b2ee100b1cd2e79e1a2fa1a0541d5
```
Where the base is the first commit that contained the initial empty nx generated workspace.
Also added in pipeline the `npm run format:check` to validate proper formatting of the changes between the branch and master. (it will continue if it fails for now due to rule discrepancy; see below)

Apparently there is a discrepancy between **stylelint** and the **nx formatter** that need to be addressed.

@ttalbot I will investigate this issue and comprise, hopefully, find a intersection between nx formatting rules & stylelint linting rulse :D


aifrim:
Now `npm run format` will ignore all `.scss` files and all `package-lock.json` files.

apps/angular-test-app/src/index.html
```diff
@@ -7,7 +7,10 @@
 
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
-    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+    <link
+      href="https://fonts.googleapis.com/icon?family=Material+Icons"
+      rel="stylesheet"
+    />
```
ttalbot: Multi-line links for that much char feels a bit overkill, isn't it?

aifrim: I think it is because the text exceeds 80 characters. I will check for 120.

aifrim: Added
```
max_line_length = 120
```
to all files except `*.md` that have it disabled

PR #22, opened 2019-09-06
feat(table): added total row/column drag drop/select event feature add demo app configurations

Ping-YUAN:


aifrim:
What new feature did you add?
Please state it for visibility and tracking.

themes/angular-theme/lib/radio-button/_radio-button-theme.scss
```diff
@@ -0,0 +1,13 @@
+@mixin uxg-radio-button-theme($theme) {
+  $primary: map-get($theme, primary);
+
+  .mat-radio-button.mat-accent {
```
ttalbot: If you are using **mat-accent** class then you cannot use **primary** color as background-color.
If you are trying to use primary as default color, then remove .mat-accent

aifrim:
> .mat-accent is for Finastra's secondary Fuchsia, .mat-primary is for Finastra's primary Violet. If you are trying to put primary as default color, remove .mat-accent from your css.

@ttalbot 
Is this a review, or can we merge?

PR #27, opened 2019-09-25
feat(repeater): added component

xmonvoisin:
uxg repeater


apps/angular-test-app/src/app/components/repeater-demo/repeater-card-example/_repeater-card-example.theme.scss
```diff
@@ -0,0 +1,12 @@
+
+
+@mixin theme-repeater-card-example($theme) {
+    $primary: map-get($theme, accent);
```
ttalbot: you're mixing primary and accent, choose between:
```SCSS
$accent: map-get($theme, accent);
// OR
$primary: map-get($theme, primary);
```

libs/angular-components/repeater/src/component-repeater/component-repeater.component.scss
```diff
@@ -0,0 +1,4 @@
+
+    uxg-component-repeater{
+        cursor: pointer;
```
ttalbot: Shouldn't this defined by the component being repeated?
Imagine repeating a card with buttons inside. You want to interact with the buttons, not necessarily the card itself, right?

libs/angular-components/repeater/src/repeater.component.scss
```diff
@@ -0,0 +1,14 @@
...
+    .cdk-virtual-scroll-orientation-horizontal {
+        .cdk-virtual-scroll-content-wrapper{
+            flex-direction: row;
+            display: flex;
+        }
+    }
+    .repeater-item{
+        display: flex;
+        &.selected{
+            background-color: red;
```
ttalbot: use theme's warn color instead of red:
```SCSS
$warn: map-get($theme, warn);
```
not sure selection should use red/warn color anyway.
Let's discuss that later 🙂

apps/angular-test-app/src/app/components/repeater-demo/repeater-card-advanced-example/repeater-card-advanced-example.component.html
```diff
@@ -0,0 +1,40 @@
...
+              <div class="strip-top"></div>        
+              <div class="investor">{{data[columnsMatcher.bankName] || 'bankName'}}</div>
+              <div class="chip">
+                <div class="chip-line"></div>
+                <div class="chip-line"></div>
+                <div class="chip-line"></div>
+                <div class="chip-line"></div>
+                <div class="chip-main"></div>
+              </div>
+              <svg class="wave" viewBox="0 3.71 26.959 38.787" width="26.959" height="38.787" fill="white">
```
aifrim: If you want custom icons, it is better to implement a custom icon registry.

libs/angular-components/tsconfig.spec.json
```diff
@@ -6,5 +6,5 @@
     "types": ["jest", "node"]
   },
   "files": ["test-setup.ts"],
-  "include": ["**/*.spec.ts", "**/*.d.ts"]
+  "include": []
```
aifrim: At one point we will test is.
Is this component part of... what part?

apps/angular-test-app/src/app/components/repeater-demo/repeater-card-advanced-example/repeater-card-advanced-example.component.html
xmonvoisin: This html bit is a copy and past from code pen, it was to illustrate we can do that.

libs/angular-components/tsconfig.spec.json
xmonvoisin: My bad

libs/angular-components/table/ng-package.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "../../../../node_modules/ng-packagr/ng-package.schema.json",
-  "dest": "../../dist/libs/angular-components/table",
```
Ping-YUAN: why we changed the dest here ?

libs/angular-components/repeater/src/repeater.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ScrollingModule as ExperimentalScrollingModule } from '@angular/cdk-experimental/scrolling';
+import { ComponentRepeaterComponent } from './component-repeater/component-repeater.component';
+
+@NgModule({
+  declarations: [RepeaterComponent, ComponentRepeaterComponent],
+  imports: [CommonModule, MatTableModule, ScrollingModule, ExperimentalScrollingModule],
+  exports: [RepeaterComponent],
+  entryComponents: [RepeaterComponent]
+})
+export class RepeaterModule {}
```
Ping-YUAN: we don't add prefix "uxg" for module ? @bcldvd

aifrim: We should avoid adding prefixes all the time since we can do this
```ts
import { SomeModule } from 'some-package';
import { SomeModule as SomeOtherModule } from 'some-other-package';
```

ngxs & ngrx both export `Store` :)

libs/angular-components/table/ng-package.json
aifrim: Maybe by mistake. Should be reverted.
@xmonvoisin

PR #35, opened 2019-10-09
feat(layout): make toolbar and breadcrumb responsive

ttalbot:


themes/angular-theme/lib/breadcrumb/_breadcrumb-base.scss
```diff
@@ -28,3 +37,30 @@
...
 }
+
+/* Mobile */
+@media only screen and (max-width: $mobile-size-up) {
+  .uxg-breadcrumb.uxg-breadcrumb-responsive {
+    li {
+      display: none;
+    }
+
+    li:not(:last-child) {
```
ttalbot: I realize that this line does not work.
should be something like
```css
li:not(:only-of-type)
```

rbordeanu:
So the current active breadcrumb link is not emphasised in any way? Only the root?
Maybe at least accent colour for current one?

What about hovering cues?

rbordeanu:
> > So the current active breadcrumb link is not emphasised in any way? Only the root?
> > Maybe at least accent colour for current one?
> > What about hovering cues?
> 
> This is not the purpose of this PR.

Fair enough :) But please consider this behaviour in a future PR.

PR #39, opened 2019-10-14
feat(bundle): properly bundle all css

aifrim:
In order for the components to contain their styles we will need to do this in the component main scss
1. import the `uxg-app-theme`
2. import the `component-theme`
3. use the component mixin

Example from the global-search component:
```scss
@import '../../../../themes/angular-theme/uxg-app-theme';

@import './global-search.theme';
@include uxg-global-search-theme($uxg-light-theme);
```

I have updated the component schematic with these new changes.

**NOTE 1:** Components that do not require theming, need not use this.
**NOTE 2:** Relative pathing to the `uxg-app-theme` did not work

**NOTE 3:** In order to allow themes we will need to refactor all the components and the theme. 
We will neeed to import `_component.theme.scss` into the `uxg-angular-theme` and build it there two times for both themes. This is a provisional solution until we make a decision on how to supply theming within the FDS.

### DIFF
--- a/apps/angular-test-app/src/app/app-routing.module.ts
+++ b/apps/angular-test-app/src/app/app-routing.module.ts
@@ -1,17 +1,28 @@
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
-import { MatCardModule, MatChipsModule, MatButtonModule, MatIconModule } from '@angular/material';
 import { RouterModule } from '@angular/router';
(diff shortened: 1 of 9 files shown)


rbordeanu:
With this change, we don't need to import them anymore in the dev app styles.css no?

aifrim:
> With this change, we don't need to import them anymore in the dev app styles.css no?

Had a bug, not it is fixed.
No need to import anything from the components to the app `style.scss`, no 🥇

PR #40, opened 2019-10-17
feat(filter-tags) Added component

adrian-apetrei:
filter-tags


adrian-apetrei:
Filter-tags preview: 
filter-tags

aifrim:
> The implementation is really with high quality . Thanks for your work : )
> In addition , there are just some details to modify or to clarify .
> 
> 1. In the right , the button " filter " should use the " actived button " , with a hover in purple ;
> 2. The color of " reset " and " clear all " is # C137A2 and align them in the right ;
> 3. When the search bar is actived, there will be a hover , the color is # 2123C0 and the opacity is 4%;
> 4. Between the search bar and the drop down menu , there is a margin-top of 4px .
>    Thanks again . Have a nice day !

The `Reset` and `Clear all` button style is from the theme. Here is the [complete list of buttons](https://material.angular.io/components/button/examples), which one should we use?

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
```diff
@@ -0,0 +1,30 @@
...
+    width: 100%;
+  }
+
+  .mat-chip.mat-standard-chip {
+    padding: 3px 8px 3px 12px;
+    margin: 0px 4px 12px 4px;
+  }
+
+  .mat-chip.mat-standard-chip .mat-chip-remove {
+    margin: 5px 0px 5px 8px;
```
ttalbot: If there's an issue with theme's margins we better have to fix the theme.

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  mat-form-field {
+    margin-top: 4px;
+  }
+  .chip-list {
+    width: 100%;
+  }
+
+  .mat-chip.mat-standard-chip {
+    padding: 3px 8px 3px 12px;
```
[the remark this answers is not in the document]
aifrim: There was extra padding in the design. We can revert this padding to default to the theme.
We chose the design as our source of truth 😄

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
aifrim: Same as this https://github.com/fusionfabric/finastra-design-system/pull/40#discussion_r337434503 comment.

From our point of view, there is no issue. We can remove this styling.

aifrim:
> > The implementation is really with high quality . Thanks for your work : )
> 
> 👍
> 
> > 1. In the right , the button " filter " should use the " actived button " , with a hover in purple ;
> 
> Filter button is not part of this component.
> 
> > 1. The color of " reset " and " clear all " is # C137A2 and align them in the right ;
> 
> This is secondary or 'accent' color.
> 
> > 1. When the search bar is actived, there will be a hover , the color is # 2123C0 and the opacity is 4%;
> 
> This is theme responsibility so you don't have to change anything at the moment
> 
> > 1. Between the search bar and the drop down menu , there is a margin-top of 4px .
> 
> 4px is `$uxg-spacing-1`

Should we add those `4px` between the search bar and the drop down?

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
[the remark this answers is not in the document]
ttalbot: cool, so remove the extra padding, I'll check if the theme is not in sync with the design 👍

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
ttalbot: please do so 👍

dxy009:
Hello ! Another thing : It is better to put a background in the part of the header " view by " . The color is # F5F5F5 . Thanks : )
Screenshot 2019-10-22 at 13 28 11

aifrim:
> Hello ! Another thing : It is better to put a background in the part of the header " view by " . The color is # F5F5F5 . Thanks : )
> Screenshot 2019-10-22 at 13 28 11

**View by** is work in progress. And not part of this PR.

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
adrian-apetrei: I removed all custom `mat-chip` styles.
Thank you!
image

PR #46, opened 2019-10-22
fix(chips-style): remove icon did not have hover style in selected state

### LINKED ISSUE #13: Chips: remove icon does not have hover style in selected state
Angular-theme's version: 0.12.0

Here is a screenshot from Chrome for mac:
Screenshot 2019-08-28 at 17 14 19

beaolivei:
fix #13

### DIFF
--- a/themes/angular-theme/lib/chips/_chips-theme.scss
+++ b/themes/angular-theme/lib/chips/_chips-theme.scss
@@ -14,10 +14,6 @@
 
     .mat-icon {
       color: mat-color($foreground, text, 0.25);
-
-      &:hover {
-        color: mat-color($primary, default);
-      }
     }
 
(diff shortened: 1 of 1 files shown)


themes/angular-theme/lib/chips/_chips-theme.scss
```diff
@@ -14,6 +14,10 @@
 
     .mat-icon {
       color: mat-color($foreground, text, 0.25);
+
+      &:hover {
+        color: mat-color($primary, default);
+      }
```
ttalbot: I think you forgot to remove hover effect on icons.

beaolivei: I removed the hover and pushed a new version without it.

PR #51, opened 2019-10-24
feat(unit-testing): added coverage collection and printing joined report

pawelleszczewicz:
added stuff created by Alex, added also html report


aifrim:
> added stuff created by Alex, added also html report

you stole my work!

pawelleszczewicz:
if i would steal it i would not mention your name :P

jest.config.js
```diff
@@ -1,10 +1,38 @@
+const esModules = ['lodash-es', '@ngrx'].join('|');
```
aifrim: we don't have any `esModules` to ignore :)

pawelleszczewicz: done

PR #58, opened 2019-11-06
feat(vector-map): add component

aifrim:
@ttalbot @beaolivei @julietteramos @julisyda @dxy009 
I created a `PaletteService` that is required mostly for chart-like components for it's output color (& design) configuration.

I added the `@studios-team` as a code owner to the default configuration of that service so that you guys can validate that we are using the proper colors / widths.

Here is the file:
```ts
// libs/angular-components/core/src/palette/palette.defaults.ts
export const PALETTE_DEFAULT_CONFIG: PaletteConfig = {
  colorScale: [
    // reversed due to a plotly bug
    [0, 'rgb(73, 54, 150)'],
    [0.1, 'rgb(73, 54, 150)'],
    [0.3, 'rgb(105, 78, 214)'],
    [0.5, 'rgb(143, 123, 225)'],
    [0.6, 'rgb(180, 166, 234)'],
    [0.7, 'rgb(217, 211, 245)'],
    [1, 'rgb(210, 210, 210)']
  ],
  vectorMap: {
    marker: {
      line: {
        color: '#fff',
        width: 0.5
      }
    }
  }
};
```

Screenshot:
image


dxy009:
Hello!
I checked all the colors which you used. About the purples, there is no problem. All the codes are right. But i don't find the code of rgb 210 210 210 in the file sketch. The most similar one is Charcoal C25 RGB 206 206 206.
Best Regards

dxy009:
Here is a picture with all the HEX colors. 
colors

aifrim:
@ttalbot 
The values are reversed due to a Plotly bug. 1 is the lowest, 0 is the biggest.
They are for the legend on the right. They need to provide a gradient from lowest to biggest. 
We can only put only one type of color nuances there, violets / purples /  fuchsias.

@dxy009 
I will change that to Charcoal C25.
But, please, in the future, put the right colors in the designs. I took the colors straight from there.

dxy009:
@aifrim Where did you take the color, plz? I dont think that we used the wrong color in the design.

aifrim:
> @aifrim Where did you take the color, plz? I dont think that we used the wrong color in the design.

image

dxy009:
@aifrim This vector-map is not the work of designers. No designer made the design or checked the colors. It is just a work of un intern. So maybe there is something wrong of the design.

dxy009:
@Ping-YUAN

aifrim:
> @aifrim This vector-map is not the work of designers. No designer made the design or checked the colors. It is just a work of un intern. So maybe there is something wrong of the design.

I removed all the work that the previous two guys did.
Started from scratch. And used the [design from Invision](https://finastra.invisionapp.com/d/main#/console/16806075/385910937/preview) as a source of truth.

dxy009:
@aifrim It is a work which is done by Heather and Chloe a month ago. At that time, we did not share the final design system with them i think. So the colors and the designs should be checked again now.

aifrim:
> @aifrim It is a work which is done by Heather and Chloe a month ago. At that time, we did not share the final design system with them i think. So the colors and the designs should be checked again now.

Checked by whom?

PR #61, opened 2019-11-13
bugfix(filter-tree):  fixed change detection on dataSource

adrian-apetrei:


apps/angular-test-app/src/app/components/filter-panel-demo/filter-panel-demo.component.ts
```diff
@@ -7,18 +8,24 @@ import { sampleFilterTree } from './filter-panel-demo.sample-data';
...
 })
 export class FilterPanelDemoComponent implements OnInit {
-  sampleData = [];
   filterExpression: string[] = [];
   groupValue: string;
   initialGroupState = 'api';
   tagsSampleData = ['Malauzai', 'North America', 'Money Movement', 'Static Data'];
   chosenTags: string[] = [];
 
+  treeSampleData$ = new Observable(observer => {
```
aifrim: You can remove this. It was only used for testing.
And go back to what was here before.

If you want, we can test this via unit testing.

PR #66, opened 2019-11-19
feat(filters): further improvements

rbordeanu:
- fix Angular problems with inheritance of @Input params from base class
- fix animation glitch


libs/angular-components/filter/filter-tree/src/filter-tree.component.ts
```diff
@@ -1,16 +1,19 @@
...
 import { MatTreeNestedDataSource } from '@angular/material';
-import { UXGFilter } from '@ffdc/uxg-angular-components/core';
 
 export interface TreeNode {
   label: string;
   children?: TreeNode[];
   parent?: TreeNode;
   isSelected?: boolean;
 }
+interface UXGFilterChanges {
```
aifrim: You could have kept the `filter.model.ts` file for this interface :)

PR #73, opened 2019-11-21
fix(angular-components): add elasticlunr as an explicit dependency

bcldvd:
to fix stackblitz's APIs limitation

### DIFF
--- a/libs/angular-components/package.json
+++ b/libs/angular-components/package.json
@@ -2,7 +2,7 @@
   "name": "@ffdc/uxg-angular-components",
   "version": "0.13.14",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


libs/angular-components/package.json
```diff
@@ -1,14 +1,16 @@
 {
   "name": "@ffdc/uxg-angular-components",
   "version": "0.13.14",
+  "dependencies": {
+    "elasticlunr": "github:weixsong/elasticlunr.js#f773890392760e0b3c68f0e65dbacb396dff297f",
```
aifrim: Well... that trailing `,` :))

Remove it 💃

bcldvd: 🙈

PR #85, opened 2019-12-06
feat(filters): filter improvements

adrian-apetrei:
Set initial state of filters in dataSource.

### DIFF
--- a/libs/angular-components/filter/filter-tree/src/filter-tree.component.scss
+++ b/libs/angular-components/filter/filter-tree/src/filter-tree.component.scss
@@ -1,5 +1,6 @@
 @import '../../../../../themes/angular-theme/uxg-app-theme';
 
+$uxg-violet: mat-color($uxg-light-primary, 500);
 $row-height: 34px;
 
 .filter-tree .child-tree,
@@ -19,4 +20,12 @@ $row-height: 34px;
   justify-content: space-between;
   min-height: $row-height;
   overflow: visible;
+}
+
+mat-checkbox {
+  &.mat-checkbox-checked, &.mat-checkbox-indeterminate {
+    .node-label {
+      color: $uxg-violet;
+    }
+  }
 }
--- a/libs/angular-components/filter/filter-tree/src/filter-tree.component.scss
+++ b/libs/angular-components/filter/filter-tree/src/filter-tree.component.scss
@@ -1,6 +1,6 @@
 @import '../../../../../themes/angular-theme/uxg-app-theme';
 
-$uxg-violet: mat-color($uxg-light-primary, 500);
+$primary: map-get($uxg-light-theme, primary);
 $row-height: 34px;
 
 .filter-tree .child-tree,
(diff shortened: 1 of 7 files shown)


libs/angular-components/filter/filter-tree/src/filter-tree.component.scss
```diff
@@ -1,5 +1,6 @@
 @import '../../../../../themes/angular-theme/uxg-app-theme';
 
+$uxg-violet: mat-color($uxg-light-primary, 500);
```
ttalbot: Instead of creating another uxg-violet, you could use the one from the theme directly by doing so:
```scss
$primary: mat-color($uxg-violet, 500);
```
or even better, extract it from the theme:
```scss
$primary: map-get($uxg-light-theme, primary);

.node-label {
  color: mat-color($primary, default);
}
```

adrian-apetrei: Done, thank you.

libs/angular-components/filter/filter-tree/src/filter-tree.component.ts
```diff
@@ -69,13 +74,18 @@ export class FilterTreeComponent implements OnInit, OnChanges {
...
   }
 
-  ngOnInit() {
-    this.dataSource.data = this.data;
-  }
-
   ngOnChanges(changes: SimpleChanges) {
     if (changes) {
       this.dataSource.data = changes.data.currentValue;
+      this.selectedData = this.getSelectedNodes(this.dataSource.data);
```
rbordeanu: you have the same code in ngOnInit as well, is this necessary?
I would think ngOnChanges is sufficient

libs/angular-components/filter/filter-tree/src/filter-tree.component.ts
```diff
@@ -69,13 +74,18 @@ export class FilterTreeComponent implements OnInit, OnChanges {
...
+      this.checkListSelection = new SelectionModel<TreeNode>(true, this.selectedData);
+      // tslint:disable-next-line: no-shadowed-variable
+      this.checkListSelection.changed.subscribe(changes => {
```
rbordeanu: this will create a new subscription every time changes occur
you should put this into a subscription variable and unsubscribe from it when you enter ngOnChanges if it is already subscribed

libs/angular-components/filter/filter-tree/src/filter-tree.component.ts
```diff
@@ -173,4 +183,20 @@ export class FilterTreeComponent implements OnInit, OnChanges {
       el.classList.add('collapsed');
     }
   }
+
+  getSelectedNodes(data: TreeNode[]) {
```
rbordeanu: this method should have a return type

libs/angular-components/filter/filter-tree/src/filter-tree.component.ts
```diff
@@ -173,4 +183,20 @@ export class FilterTreeComponent implements OnInit, OnChanges {
...
+    let result: TreeNode[];
+    result = data.reduce((acc, node) => {
+      if (node.isSelected) acc.push(node);
+      if (node.children) {
+        const childNodes = node.children.reduce((accu, subNode) => {
```
rbordeanu: this reducer feels a little extra verbose, I think you can just do
`childNodes = node.children.filter(n => n.isSelected)` no?

PR #87, opened 2019-12-10
fix(table): fix table selection checking method to deep check

Ping-YUAN:


libs/angular-components/table/src/table.component.ts
```diff
@@ -396,4 +395,15 @@ export class TableComponent implements OnInit, OnDestroy, OnChanges {
...
     }
   }
+
+  isRowSelected(row) {
+    return this.getSelectedIndex(row) > -1 ? true : false;
+  }
+
+  getSelectedIndex(row) {
+    const rowIdx = this.selections.findIndex(item => {
+      return isEqual(item, row);
```
aifrim: Won't `Object.is` provide the same level of confidence?

If so, I propose we remove `lodash` as a dep.

@bcldvd ?

libs/angular-components/table/src/table.component.ts
```diff
@@ -14,7 +14,7 @@ import { MatTable, PageEvent, MatCheckbox } from '@angular/material';
 
 import { UxgColumn, UxgSort, UxgPage, UxgColumnType, UxgTableSelectEvent, UxgDefaultPaging } from './table.models';
 import { CdkDropList, moveItemInArray } from '@angular/cdk/drag-drop';
-
+import { isEqual } from 'lodash';
```
aifrim: If we chose to keep this, can you
```ts
import { isEqual } from 'lodash/isEqual';
```

So that only `isEqual` is loaded to memory, not the entire `lodash` library :)

And also add it to `libs/angular.components/package.json` as a `peerDependencies`.

libs/angular-components/table/src/table.component.ts
bcldvd: Well since [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) is not compatible with IE, will the CLI provide the correct polyfills for it ?

Ping-YUAN: keep lodash

aifrim: True that, but in IE 11 like all browsers references are kept.
And in that case it should be `return item === row`.

xmonvoisin: Issue reside is the fact that when pagination is on (and asynchronous) when going to page 1 then page 2 and going back page 1, page 1 data is not the same reference as first page 1.

libs/angular-components/table/src/table.component.ts
```diff
@@ -14,7 +14,7 @@ import { MatTable, PageEvent, MatCheckbox } from '@angular/material';
 
 import { UxgColumn, UxgSort, UxgPage, UxgColumnType, UxgTableSelectEvent, UxgDefaultPaging } from './table.models';
 import { CdkDropList, moveItemInArray } from '@angular/cdk/drag-drop';
-
+import { isEqual } from 'lodash/isEqual';
```
aifrim: Do not use `lodash`, but use `lodash-es` and this will (should) work.
Update both root package.json and local components peerDeps.

libs/angular-components/table/src/table.component.ts
```diff
@@ -396,4 +395,15 @@ export class TableComponent implements OnInit, OnDestroy, OnChanges {
...
     }
   }
+
+  isRowSelected(row) {
+    return this.getSelectedIndex(row) > -1 ? true : false;
+  }
+
+  getSelectedIndex(row) {
+    const rowIdx = this.selections.findIndex(item => {
+      return item === row || isEqual(item, row);
```
aifrim: Since references are not kept, just use isEqual.

libs/angular-components/table/package.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "@ffdc/uxg-angular-components/table",
   "version": "0.13.21",
-  "author": "pingyuan <ping.yuan@finastra.com>"
+  "author": "pingyuan <ping.yuan@finastra.com>",
+  "peerDependencies": {
```
aifrim: Not here, this should be added in `/libs/angular-components/package.json`

PR #88, opened 2019-12-10
feat(wizard): new wizard component

rbordeanu:


libs/angular-components/wizard/src/services/wizard-navigation.service.ts
```diff
@@ -0,0 +1,221 @@
...
+export class WizardNavigationService implements OnDestroy {
+  public previousButtonSubscription: Subscription;
+
+  public nextButtonSubscription: Subscription;
+
+  public doneButtonSubscription: Subscription;
+
+  public cancelButtonSubscription: Subscription;
+
+  private _currentChange = new Subject<UxgWizardPage>();
```
aifrim: Due to the existence of access modifiers, we no longer need to prefix with `_`.

The only exception is when you use getters and setters.
```ts
private _data: T;
public get data(): T { return this._data; }
public set data(value: T) { this._data = value; }
```

And even then, in the class methods, you should use the `this.data`, most of the time.

libs/angular-components/wizard/src/services/wizard-navigation.service.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (!previousPage) {
+      return;
+    }
+
+    this._movedToPreviousPage.next(true);
+
+    this.currentPage = previousPage;
+  }
+
+  public goTo(pageOrId: any) {
```
aifrim: `public goTo(pageOrId: any)` to `public goTo(pageOrId: UxgWizardPage | string)`

libs/angular-components/wizard/src/wizard-button/wizard-button.component.ts
```diff
@@ -0,0 +1,138 @@
+import { Component, OnInit, Input, Output, EventEmitter, ViewEncapsulation } from '@angular/core';
+import { WizardNavigationService } from '../services/wizard-navigation.service';
+import { ButtonHubService } from '../services/button-hub.service';
+
+export const DEFAULT_BUTTON_TYPES: any = {
```
aifrim: Isn't it better to have it as a `enum`?
Since enum is both an object and a data type?
```ts
export enum WizardButtonType {
  Cancel = 'cancel',
  Previous = 'previous',
  Next = 'next',
  Done = 'done',
  Custom = 'custom'
};
```

Or an aggregated string type:
```ts
export type WizardButtonType = 'cancel' | 'previous' | 'next' | 'done' | 'custom';
```

And maybe create a `.models.ts` file where we can add all the necessary custom types used in this component.

libs/angular-components/wizard/src/wizard-button/wizard-button.component.ts
```diff
@@ -0,0 +1,138 @@
...
+@Component({
+  selector: 'uxg-wizard-button',
+  templateUrl: './wizard-button.component.html',
+  host: {
+    '[class.uxg-wizard-button]': 'true',
+    '[class.auto-margin]': 'isCancel || isCustom',
+    '[attr.aria-hidden]': 'isHidden'
+  }
+})
+export class UxgWizardButton implements OnInit {
```
aifrim: Why implement `OnInit` if the `ngOnInit` is empty?

libs/angular-components/wizard/src/wizard-page/wizard-page.component.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  private _disabled = false;
+
+  public get disabled(): boolean {
+    return this._disabled;
+  }
+
+  @Input('uxgWizardPageDisabled')
+  public set pageDisabled(val: boolean) {
+    const valBool = !!val;
```
aifrim: If `val` is `boolean` why create `valBool` that is `boolean`

libs/angular-components/wizard/src/wizard-page/wizard-page.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (valBool !== this._disabled) {
+      this._disabled = valBool;
```
aifrim: Maybe rename `this._disabled` to `_pageDisabled`.
The `_disabled` property is set via `set pageDisabled` and get via `get disabled` (confusing 😞)

libs/angular-components/wizard/src/wizard.component.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  private listenForPageChanges(): Subscription {
+    return this.navService.currentPageChange.subscribe(page => {
+      this.currentPageId = this.pageCollection.getPageIndex(page);
+      this.currentPageChange.emit();
+    });
+  }
+}
+
+@Component({
```
aifrim: Can we have one component per file?

libs/angular-components/wizard/src/wizard.module.ts
```diff
@@ -0,0 +1,23 @@
...
+import { UxgWizardButton } from './wizard-button/wizard-button.component';
+import { UxgWizardPageDescription } from './wizard-page/wizard-page-description.directive';
+import { UxgWizardPageTitle } from './wizard-page/wizard-page-title.directive';
+import { UxgWizardPage } from './wizard-page/wizard-page.component';
+import { UxgWizard, UxgWizardTitle } from './wizard.component';
+
+@NgModule({
+  imports: [CommonModule, FlexLayoutModule, MatIconModule, MatButtonModule, MatDividerModule],
+  declarations: [
+    UxgWizard,
```
aifrim: We should suffix all with `Component`, no?
@bcldvd

apps/angular-test-app/src/app/components/wizard-demo/wizard-demo.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'ffdc-wizard-demo',
+  templateUrl: './wizard-demo.component.html',
+  styleUrls: ['./wizard-demo.component.scss']
+})
+export class WizardDemoComponent implements OnInit {
```
aifrim: We should not `implement OnInit` if we do not have anything in `ngOnInit`

libs/angular-components/wizard/src/wizard-button/wizard-button.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    if (this.isDone && !nav.currentPageIsLast && !this.alwaysShow) {
+      return hidden;
+    }
+
+    return !hidden;
+  }
+
+  public get _disabledAttribute(): string | null {
```
aifrim: remove `_` prefix since this is public.

libs/angular-components/wizard/src/wizard.module.ts
bcldvd: da

libs/angular-components/wizard/src/_wizard.theme.scss
```diff
@@ -0,0 +1,182 @@
...
+  $accent: map-get($theme, accent);
+  $background: map-get($theme, background);
+
+  $wizard-background-color: mat-color($background, background);
+  $wizard-step-active-color: mat-color($accent, default);
+  $wizard-step-done-color: map-get($uxg-grass, 100);
+  $wizard-timeline-color: map_get($mat-grey, 300);
+
+  $wizard-dark-blue: #060546;
+  $wizard-step-description-color: map-get($uxg-violet, 100);
```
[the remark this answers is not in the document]
ttalbot: Forget about it, how we made the palette is confusing, you're using 100 hue and default is 500 so your code is totally valid but for secondary colors 100 means 100% opacity. It lacks consistency.

libs/angular-components/wizard/src/services/wizard-navigation.service.ts
rbordeanu: these ARE getters and setters

libs/angular-components/wizard/src/wizard-page/wizard-page.component.ts
rbordeanu: was doing type coercion initially but I changed the parameters type after

libs/angular-components/wizard/src/wizard-page/wizard-page.component.ts
rbordeanu: renamed everything to disabled

rbordeanu:
> Looks good, just noticed that you're mixing theme (colors) with sizes and typography in the SCSS. Another remark is that you use both Flexbox in CSS and angular flex-layout in HTML... It would be better if you could choose. I personally think layout belong to styling.

I refactored the CSS a bit, please have another look.

PR #93, opened 2019-12-17
feat(table): custom cell template

Ping-YUAN:
developer can pass a template to table component to customize the table column display


libs/angular-components/table/src/table.component.scss
```diff
@@ -80,6 +80,10 @@ uxg-table {
 
         thead {
             tr {
+                &.uxg-table-no-header{
+                    display: none !important;
```
aifrim: I'd try to find another solution to disable the header, instead of using `!important`

@ttalbot

PR #100, opened 2020-01-07
fix(table) fix set selectionKeys

xmonvoisin:
Fix the set selectionkeys to inject selection


libs/angular-components/table/src/table.component.ts
```diff
@@ -83,9 +83,10 @@ export class TableComponent implements OnInit, OnDestroy, OnChanges {
       }
     }
     if (!this.singleSelect && this.multiSelect) {
-      selectedIndex.forEach(selectionIndex => {
-        if (this.isRowSelected(this.data[selectedIndex[0]])) {
-          this.selections.push(this.data[selectionIndex]);
+      this.selections = [];
```
aifrim: can we not reassign properties?
`this.selections.length = 0` if you want to clean it

PR #103, opened 2020-01-09
feat(theme): refactor to separate uxg-core and uxg-material-theme

ttalbot:
⚠️ BE CAREFUL - IT COULD BREAK THE WAY YOU IMPORT THE THEME ⚠️

2 choices:
- use one of the provided files (all-theme, light or dark) and you're good to go.
- do all the setup manually so that you can prevent calling uxg-core twice

now you call uxg-core() for all base styles (margin / padding / witdh...) at app level.
Then you call uxg-material-theme to generate the actual theme (the css applying colors)

I refactored the components so that their theme styles are only generated at app level where you create the theme.


aifrim:
I have added Adrian to review changes on filter-tags. Apparently there are problems.
He will create an issue :)

PR #104, opened 2020-01-13
fix(toggle-button): fix typography and icon margin

ttalbot:
Screenshot 2020-01-13 at 10 03 46

### DIFF
--- a/themes/angular-theme/lib/button-toggle/_button-toggle-base.scss
+++ b/themes/angular-theme/lib/button-toggle/_button-toggle-base.scss
@@ -2,7 +2,6 @@
 @import '../icon/icon-base';
 
 $button-toggle-margin: $uxg-spacing-2;
-$button-toggle-dense-height: 30px;
 $button-toggle-icon-dense: 18px;
 
 $filtering-button-toggle-radius: $uxg-spacing-1;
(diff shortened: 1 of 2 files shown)


adrian-apetrei:
Hello, one small thing I noticed on disabled buttons:
image
Can you please remove the outline and maybe change color on disabled buttons?

aifrim:
> > Hello, one small thing I noticed on disabled buttons:
> > image
> > Can you please remove the outline and maybe change color on disabled buttons?
> 
> I updated the disabled state.
> Not sure how to handle the blue outline properly, it's the focus state provided by the browser, we cannot just remove it since it's an accessibility feature. I'll think about it. Thanks 👍

Disabled buttons should not be focusable.

rbordeanu:
> > > > Hello, one small thing I noticed on disabled buttons:
> > > > image
> > > > Can you please remove the outline and maybe change color on disabled buttons?
> > > 
> > > 
> > > I updated the disabled state.
> > > Not sure how to handle the blue outline properly, it's the focus state provided by the browser, we cannot just remove it since it's an accessibility feature. I'll think about it. Thanks 👍
> > 
> > 
> > Disabled buttons should not be focusable.
> 
> Well it appears that it is not obvious that disabled buttons should not be focusable.
> Speaking of visually impaired users with screen reader :
> [...] _If disabled buttons are not focusable with TAB, our user would occasionally miss that the button ever exists, and they will never find it._
> I did not test this personally, but sounds legit.
> I can only affirm that this focus border is default angular-material behaviour.

But in Material, the disabled button is NOT focusable with TAB anyway, the ugly outline only appears on direct click on the button...

PR #111, opened 2020-01-20
feat(filter-group) added filter-group and filter-toggle components

adrian-apetrei:
filter-group-v2

https://finastra.invisionapp.com/d/main#/console/18479837/399554321/preview


dxy009:
Hi Adrian,

Thanks soooo much for your work. I really like it : )
Here are my comments:
1. We changed the style of text fields in all the design. Now we use only the filled ones. So maybe we should change the one below ' Tags ' . If you have any questions for this, you could ask Julien for more details.
2. There is also a text field in the ' Name of filter ' . By default, there is a light purple inside. And when it is activated, the color turns to medium purple.
3.  In my opinion, maybe it is better to separate the ' Time Frame ' and the others. So in the design, i put a dark divider. 
4. For all the big titles, we use the typo ' FuturaEF Bold ' . 
5. For the action ' Cancel ' , i think when we click it, the choices will not be saved. The ' Apply ' plays this role. 
6. The big action with dropdown, by default, its outline is light grey. And for the activated state it turns to purple. 

That is all from me. Thanks again for your excellent work!
Have a nice day!
Xinyi

dxy009:
Hi !
Maybe we should add Chloe also for the review ?
Thanks : )
Xinyi

aifrim:
> Hi !
> Maybe we should add Chloe also for the review ?
> Thanks : )
> Xinyi

How should we add Chloe for review?

libs/angular-components/filter/filter-group/src/_filter-group.theme.scss
```diff
@@ -0,0 +1,57 @@
...
+
+    &.active {
+      .mat-expansion-panel-header {
+        border-color: mat-color($primary, default);
+      }
+    }
+  }
+
+  .expansion-panel-header {
+    background-color: map_get($mat-grey, 100);
```
ttalbot: For this one to be themable you should consider:
background-color: mat-color($background, background);
Or
`background-color: mat-color($background, status-bar);`
Or change the opacity
`background-color: mat-color($background, status-bar, 0.5);`
if it's too dark for your needs

libs/angular-components/filter/filter-group/src/_filter-group.theme.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .filter-group-dialog-header {
+    background-color:  map_get($mat-grey, 100);
```
ttalbot: Same here, use a color from $theme, probably same than the one you'll choose for .expansion-panel-header

libs/angular-components/filter/filter-group/src/_filter-group.theme.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+@mixin uxg-filter-group-typography($config) {
+  .mat-expansion-panel-header {
+    .uxg-filter-group-header {
+      .selectedFilters {
+        @include mat-typography-level-to-styles($config, h6);
```
ttalbot: If you need smaller font-size than h6 you should probably not use h6 and Futura, for consistency.

adrian-apetrei: Changed it to `subtitle-3` to match design style.

libs/angular-components/filter/filter-group/src/filter-group.component.html
```diff
@@ -0,0 +1,83 @@
...
+      >
+        <mat-icon *ngIf="!filterGroupPanel.expanded">arrow_drop_down</mat-icon>
+        <mat-icon *ngIf="filterGroupPanel.expanded">arrow_drop_up</mat-icon>
+      </button>
+    </div>
+  </mat-expansion-panel-header>
+
+  <div class="expansion-panel-header">
+    <div class="header-title">
+      <span>Filter by</span>
```
ttalbot: I think it would be cool to see semantic level in the HTML, with:
```SASS
<h6 class="uxg-h6">Filter by</h6>
```
It could really simplify your code, by removing a level of div here, removing the font style in uxg-filter-group-typography and removing a level of flex in filter-group.component.scss.
What do you think?

libs/angular-components/filter/filter-group/src/filter-group.component.html
```diff
@@ -0,0 +1,83 @@
...
+        </mat-chip>
+      </mat-chip-list>
+    </div>
+
+    <button *ngIf="savedFilters.length" mat-button color="accent" (click)="updateFilter()">Update</button>
+    <button *ngIf="savedFilters.length" mat-button color="accent" (click)="clearSelection()">Clear</button>
+    <mat-divider vertical="true"></mat-divider>
+    <button mat-stroked-button color="accent" (click)="openDialog()">Save as</button>
+    <button mat-stroked-button color="accent" (click)="filterGroupPanel.toggle(); close()">Cancel</button>
+    <button mat-stroked-button color="accent" (click)="applyFilters()">Apply</button>
```
ttalbot: Primary action would benefit from being a mat-flat-button

libs/angular-components/filter/filter-group/src/filter-group.component.html
adrian-apetrei: I agree with you, but I still had to add some lines of code in filter-group.component.scss to align the h6 with the button.

Thank you!

libs/angular-components/filter/filter-group/src/filter-group.component.scss
```diff
@@ -0,0 +1,81 @@
...
+    width: 100%;
+    display: flex;
+    align-items: center;
+    cursor: default;
+
+    .fill-remaining-space {
+      flex: 1 1 auto;
+    }
+
+    button {
```
ttalbot: avoid styling element directly because it's so strong it won't be possible to override it without `!important`.
Instead use classes 👍

libs/angular-components/filter/filter-group/src/filter-group.component.scss
```diff
@@ -0,0 +1,81 @@
...
+      min-width: fit-content;
```
ttalbot: fit-content is [not supported by IE and Edge](https://caniuse.com/#search=min-width%3A%20fit-content)

aifrim:
Please rebase with master to use Angular 9 and the new way of importing this.

PR #112, opened 2020-01-21
feature(components): bundle _*.theme.scss files into the components package

aifrim:
Now all `_*.theme.scss` files will be bundled alongside the components.

Usage in projects:
```scss
@import '~@ffdc/uxg-angular-components/global-search/global-search.theme';
@include uxg-global-search-theme($uxg-light-theme);
```

Other changes:
Added style include paths meaning that for the test app you can:
```scss
@import 'global-search/src/global-search.theme';
```
And within the component theme:
```scss
@import 'angular-theme/lib/core/layout/breakpoints';
```


angular.json
```diff
@@ -17,6 +17,9 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
+            "stylePreprocessorOptions": {
+              "includePaths": ["./themes/", "./libs/angular-components"]
```
code2nguyen: With this option we lost Ctrl +click to go to scss file from IDE

tools/devkit/builders/components/components-builder.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return new Observable(subscriber => {
+    execute(
+      {
+        project: options.project,
+        tsConfig: options.tsConfig
+      },
+      context
+    ).subscribe(result => {
+      subscriber.next(result);
```
code2nguyen: You should not call next if your copy task is not completed. Maybe, have side effect here

PR #118, opened 2020-01-22
fix(theme): flex: 1 removed from default toggle button

beaolivei:


aifrim:
Please, in the future, name your branches according to an extended git flow naming convention to express what this change is
- feature / feat
- bugfix / fix
- test / e2e
- doc (documentation) 

And, regardless of the prefix (see above), your branch name should describe what is done.

beaolivei:
@rbordeanu if it works best with what has already been done in other parts of the code, I will just change it in .uxg-filtering-toggle-group on the website. So, I am closing this pull request.

@aifrim thanks for the heads up. I had forgotten but now it is duly noted here.

PR #123, opened 2020-01-27
feat(vector-map): changed projection type, so it better fills the uxp component

achilehero:
New look in UXP component:
image

vs current look:

image


rbordeanu:
Well this is distorting the 2d projection, so not really what we wanted :(

aifrim:
> Well this is distorting the 2d projection, so not really what we wanted :(

I really think that Plotly, for FDS, is not the way to go.

bcldvd:
> > Well this is distorting the 2d projection, so not really what we wanted :(
> 
> I really think that Plotly, for FDS, is not the way to go.

Which library would you suggest ?

achilehero:
Changed the projection type and legend size:

aifrim:
> > > Well this is distorting the 2d projection, so not really what we wanted :(
> > 
> > 
> > I really think that Plotly, for FDS, is not the way to go.
> 
> Which library would you suggest ?

D3

libs/angular-components/vector-map/src/vector-map.component.scss
```diff
@@ -18,47 +16,40 @@
...
     display: inline-block;
     width: 10px;
     height: 10px;
     margin-top: 5px;
   }
-
   .text {
     display: inline-block;
     line-height: 20px;
+    font-size: 12px;
```
rbordeanu: can we use one of the existing typography styles, instead of hardcoding font-size and line-height?

libs/angular-components/vector-map/src/vector-map.component.scss
```diff
@@ -18,47 +16,40 @@
...
     @media only screen and (max-width: $mobile-size-up) {
       display: inline-flex;
       margin-right: $uxg-spacing-3;
     }
   }
-
   .color {
     border-radius: 50%;
     display: inline-block;
     width: 10px;
```
rbordeanu: can we add all numerical values as variables at the top, in material fashion

libs/angular-components/vector-map/src/vector-map.component.scss
achilehero: Fixed

libs/angular-components/vector-map/src/vector-map.component.scss
achilehero: moved sizes in variables

aifrim: `uxg-legend` class

achilehero: ```uxg-legend``` class has font size 10px, while all other components use 12px in the legend. Should we change the uxg-legend mixin to a 12px font and use it here? Or we just use it like I did? @ttalbot , @rbordeanu opinions?

ttalbot: Yes, legends should use uxg-legend which was specified with a 10px font-size in the design system UI Kit. Other components should use that too.

ttalbot: Well, sorry, I just double checked, looks like in the UI Kit some legends uses uxg-body-3 when it's 12px and some uxg-legend when it's 10px.

libs/angular-components/vector-map/src/vector-map.component.scss
```diff
@@ -1,11 +1,11 @@
...
-
+@import 'angular-theme/lib/core/typography/typography';
+$legend-bullet-size: 10px;
 .vector-map {
   display: block;
   position: relative;
   height: 100%;
-
   padding: $uxg-spacing-3;
   background-color: #fff;
```
ttalbot: use $theme card background if you want the component to be someday themable
```SCSS
$background: map-get($theme, background);

.vector-map {
  background-color: map-get($background, card);
}
```

libs/angular-components/vector-map/src/vector-map.component.html
```diff
@@ -30,7 +30,7 @@ <h3 class="uxg-h3">{{ title }}</h3>
           <ol>
             <li *ngFor="let item of legend">
               <span class="color" [ngStyle]="{ background: item.color }"></span>
-              <span class="text">{{ item.text }}</span>
+              <span class="text uxg-caption">{{ item.text }}</span>
```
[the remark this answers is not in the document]
achilehero: That's how I had it initially, but thought ```uxg-caption``` would be more appropriate. I'll switch it back

libs/angular-components/vector-map/src/vector-map.component.scss
```diff
@@ -1,11 +1,11 @@
 @import 'angular-theme/lib/core/style/spacing';
 @import 'angular-theme/lib/core/layout/breakpoints';
-
+@import 'angular-theme/lib/core/typography/typography';
```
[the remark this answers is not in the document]
achilehero: Cleaned

libs/angular-components/vector-map/src/vector-map.component.scss
achilehero: Created a small theme for vector-map

libs/angular-components/vector-map/src/_vector-map.theme.scss
```diff
@@ -0,0 +1,13 @@
+@import '~@angular/material/theming';
```
rbordeanu: I don't think this import is needed

libs/angular-components/vector-map/src/vector-map.component.scss
aifrim: Since there is an `uxg-legend` we should use it.
Should @achilehero update the `uxg-legend` to be 12px or leave it at 10px?

ttalbot: uxg-legend is 10px, so in this case it's correct to use uxg-body-3 👍

PR #124, opened 2020-01-27
feat(table): add optional emiter to cell template

xmonvoisin:
You can create custom template cell in the table. We needed a way to output event if we wanted to trigger anything outside of this template.


aifrim:
Please rebase with master to use Angular 9 and the new way of importing this.

PR #125, opened 2020-01-29
fix(icon-button): fixed icon-button behaviour when clicked in IE11

beaolivei:
mat-icon-button, when clicked in IE11, was displaying a crocked display. Now the icon is centered again.
icon-button-issue-ie


beaolivei:
new solution added, deleting the important tag and adjusting the wrapper to work with the icon dense buttons

PR #129, opened 2020-02-10
chore(update-deps): upgrade to Angular 9

rbordeanu:


aifrim:
We need to change the pipeline due to [this](https://angular.io/guide/creating-libraries#publishing-your-library)
> It is not recommended to publish Ivy libraries to NPM repositories.

And we need to set up a production build for the components.
> An unhandled exception occurred: Configuration 'production' is not set in the workspace.

Apparently the change is not that... simple. I gotta investigate a bit.

PR #131, opened 2020-02-12
feat(app): add theme in test app

ttalbot:
Theme is back in the test app, It will give a quick overview of UI Elements states when applying fixes or new features.


aifrim:
Will this PR add changes to the theme itself? Or only showcase it on the test app?

apps/angular-test-app/src/app/components/foundations-demo/foundations-demo.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'foundations-demo',
```
aifrim: Prefix needs to be `ffdc-`

PR #133, opened 2020-02-13
feat(app): enable dark theme in test app

ttalbot:
Add a checkbox to switch to dark theme.
It reveals some components are not theme ready.

Screenshot 2020-02-13 at 14 33 28

### DIFF
--- a/apps/angular-test-app/src/app/app.component.ts
+++ b/apps/angular-test-app/src/app/app.component.ts
@@ -11,7 +11,12 @@ export class AppComponent {
   title = '';
   dark = false;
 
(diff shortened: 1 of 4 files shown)


apps/angular-test-app/src/app/app.component.ts
```diff
@@ -8,12 +9,32 @@ import { Router, NavigationEnd, ActivatedRoute } from '@angular/router';
...
       }
     });
   }
+
+  toggleTheme() {
+    const darkThemeClass = 'uxg-dark-theme';
+
+    this.dark = !this.dark;
+
+    if (this.dark) {
```
aifrim: https://stackoverflow.com/a/52536062/3856582

A better way to add / remove classes

PR #135, opened 2020-02-14
feat(skeleton-text): New component uxg-skeleton-text

bcldvd:


apps/angular-test-app/src/styles.scss
```diff
@@ -29,6 +30,7 @@ $uxg-typography: uxg-typography-config();
     @include uxg-vector-map-theme($theme);
     @include uxg-chart-theme($theme);
     @include uxg-wizard-theme($theme);
+    @include uxg-skeleton-text-colors($theme);
```
ttalbot: It would be nice to use the pattern uxg-MY-COMPONENT-theme

bcldvd: Oki, will make a PR to devkit then !

PR #141, opened 2020-02-17
fix: upgrade angular 9

code2nguyen:


libs/angular-components/vector-map/src/vector-map.component.ts
```diff
@@ -49,9 +48,9 @@ export class VectorMapComponent implements OnInit, OnDestroy, OnChanges {
   @Output() viewChange = new EventEmitter<VectorMapView>();
 
   countries: VectorMapCountry[] = [];
-  data: Partial<Plotly.Data>[] = [];
-  layout: Partial<Plotly.Layout> = {};
-  config: Partial<Plotly.Config> = {};
+  data: any[] = [];
```
aifrim: Why remove this?
I have added them for intelisense :(

package.json
```diff
@@ -8,6 +8,7 @@
     "clean": "rimraf dist coverage junit.xml",
     "start": "ng serve",
     "build": "ng build",
+    "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points",
```
aifrim: Who does this help?
`ng b` does it when you run it.

Why do we need it as a post install?

libs/angular-components/vector-map/src/vector-map.component.ts
code2nguyen: to remove this warning
Capture d’écran 2020-02-18 à 10 16 34

package.json
code2nguyen: To speed up ng build on CI...https://angular.io/guide/ivy#speeding-up-ngcc-compilation

PR #146, opened 2020-02-25
feat(powerBI): adding power BI json theme file and readme

beaolivei:
This JSON file adapts Power BI's style accordingly to Finastra's Design System, this way the internal stakeholders will be able to produce reports that fall into the brand guidelines.


aifrim:
@ttalbot 
Do we publish this? How will this be consumed?

beaolivei:
@aifrim I added also a mention to the Power BI Theme in the main READ.ME

PR #147, opened 2020-02-26
fix: can not compile with ivy

code2nguyen:


libs/angular-components/core/src/lazyload-script.ts
```diff
@@ -0,0 +1,43 @@
+import { Injectable, Inject } from '@angular/core';
```
bcldvd: This all seems a bit cumbersome. Can't we make a PR to plotly to fix whatever needs fixing ?

code2nguyen: I need it to fix Plotly.......Maybe we can have another solution.....but for now, I dont have any other.

aifrim: I'd just remove Plotly since it is cumbersome to use.
We need:
1. Map projection using Mercador
2. Pie / Donut chart
3. Line chart
4. Bar chart

And apparently Plotly is giving us more problems that it's worth.
By using D3 we can enfore FDS better within the charts.

code2nguyen: Yes....we should use only D3

PR #154, opened 2020-03-04
bugfix(form field icon): icon centralized in input field

beaolivei:
The icon is centralized now. This pull request also starts adapting the input to its new fill style.


themes/angular-theme/lib/input/_input-base.scss
```diff
@@ -1,4 +1,13 @@
-.mat-form-field-prefix,
-.mat-form-field-suffix {
-  align-self: flex-end;
+.mat-form-field-prefix {
+  .mat-icon {
+    margin-right: 12px;
```
rbordeanu: We shouldn't impose a margin on the mat-icon prefix.
Let the implementer choose if he wants one.

themes/angular-theme/lib/input/_input-base.scss
```diff
@@ -1,4 +1,13 @@
...
+  }
+}
+
+.mat-form-field-appearance-legacy {
+  .mat-form-field-wrapper .mat-form-field-infix {
+    .mat-input-element {
+      padding: 4px 0;
```
rbordeanu: This causes the input text to not be on the same alignment with the placeholder:
Screenshot 2020-03-09 at 00 47 16
Screenshot 2020-03-09 at 00 47 30

themes/angular-theme/lib/input/_input-base.scss
```diff
@@ -1,4 +1,13 @@
-.mat-form-field-prefix,
-.mat-form-field-suffix {
-  align-self: flex-end;
```
rbordeanu: If you remove the 4px added padding (see the other comment), we should keep this setting, but only for legacy appearance.

themes/angular-theme/lib/input/_input-base.scss
beaolivei: this was removed as well. Thanks for pointing it out 👍

themes/angular-theme/lib/input/_input-base.scss
beaolivei: Absolutely right 👍

themes/angular-theme/lib/icon/_icon-base.scss
```diff
@@ -6,10 +6,10 @@ $icon-size-base: $uxg-spacing-4;
 $icon-size-dense: 18px;
 
 @mixin icon-size($icon-size) {
-  width: $icon-size;
-  height: $icon-size;
-  font-size: $icon-size;
-  line-height: $icon-size;
+  width: $icon-size !important;
```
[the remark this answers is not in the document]
rbordeanu: it's ok, they were there from the beginning :) not a recommended pattern but sometimes it's needed...

PR #155, opened 2020-03-06
update: upgrade angular 9.0.5

code2nguyen:


angular.json
```diff
@@ -25,6 +25,7 @@
             "main": "apps/angular-test-app/src/main.ts",
             "polyfills": "apps/angular-test-app/src/polyfills.ts",
             "tsConfig": "apps/angular-test-app/tsconfig.app.json",
+            "aot": true,
```
rbordeanu: ehm, we need this, why would you want to remove it?

PR #163, opened 2020-03-12
bugfix(filter-tree): fix collapse/expand animation

rbordeanu:


libs/angular-components/filter/filter-tree/src/filter-tree.component.scss
```diff
@@ -1,6 +1,6 @@
 @import 'angular-theme/lib/core/style/spacing';
 
-$row-height: 34px;
+$row-height: 32px;
```
ttalbot: 32px is the equivalent of `$uxg-spacing-3 * 2`

PR #164, opened 2020-03-15
bugfix(filters): improve themes + fix selection logic

rbordeanu:


adrian-apetrei:
Hi Razvan,
   It looks very nice now!
   I found one issue, when the action panel is used and you save a filter, the saved filters list is updated but not reflected in the template until a new change is made and one more thing I noticed is that the preselected filters are applied even tho the autoApply is false. 

Thank you

libs/angular-components/filter/filter-group/src/_filter-group.theme.scss
```diff
@@ -2,49 +2,75 @@
   $primary: map-get($theme, primary);
   $foreground: map-get($theme, foreground);
   $background: map-get($theme, background);
-  
-  .uxg-side-badge .mat-badge-content {
-    color: mat-color($primary, default);
-    background-color: mat-color($primary, lighter);
-    border-color: mat-color($primary, default, 0.15);
-  }
+  $uxg-filter-group-border-radius: 4px;
```
ttalbot: use $uxg-spacing-1

libs/angular-components/filter/filter-group/src/filter-group.component.html
```diff
@@ -1,11 +1,15 @@
 <mat-expansion-panel
   #filterGroupPanel
-  class="uxg-filter-group"
+  class="uxg-filter-group mat-elevation-z0"
```
ttalbot: Could you use SCSS mixin instead. I see it is already used for .mat-expansion-panel-content, just to stay consistent.

libs/angular-components/filter/filter-group/src/filter-group.component.scss
```diff
@@ -1,11 +1,7 @@
 @import 'angular-theme/lib/core/style/spacing';
 @import '~@angular/material/theming';
 
-.uxg-filter-group.mat-expansion-panel:not(.mat-expanded) {
-  &.active {
-    @include mat-elevation(0);
-  }
-}
+$uxg-filter-group-border-radius: 4px;
```
ttalbot: already declared in `_filter-group.theme.scss`. remove it from theme. use `$uxg-spacing-1`.

libs/angular-components/filter/filter-group/src/filter-group.component.html
```diff
@@ -1,11 +1,15 @@
...
   hideToggle="true"
   [expanded]="expanded"
-  [class.active]="checkIsActive()"
+  [class.active]="isActive"
 >
-  <mat-expansion-panel-header [expandedHeight]="'48px'" (click)="filterGroupPanel.toggle()">
+  <mat-expansion-panel-header
+    [expandedHeight]="headerHeight"
+    [collapsedHeight]="headerHeight"
```
ttalbot: expandedHeight and collapsedHeight have the same height? is it normal?

libs/angular-components/filter/filter-group/src/filter-group.component.ts
```diff
@@ -36,13 +40,18 @@ export class FilterGroupComponent implements AfterViewInit, OnDestroy {
...
+
   subscriptions: Subscription[] = [];
   selectedData: FilterGroupComponentData[] = [];
   savedFilters: SavedFilter[] = [];
   existingFilterNames: string[] = [];
   initialState: FilterGroupComponentData[] = [];
   activeFilter: FilterGroupComponentData[] = [];
   isActive = false;
+
+  headerHeight = '40px';
```
ttalbot: Could this be handled by SCSS?

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
```diff
@@ -1,26 +1,54 @@
 @mixin uxg-filter-tags-theme($theme) {
   $primary: map-get($theme, primary);
   $filter-tags-background-color: mat-color($primary, default, .12);
+  $filter-tags-option-height: 32px;
```
ttalbot: If possible this should go in filter-tags.component.scss

libs/angular-components/filter/filter-group/src/filter-group.component.html
rbordeanu: Yes we'd like to keep header unaffected by the opening/closing of the filter.

libs/angular-components/filter/filter-group/src/filter-group.component.ts
rbordeanu: Screenshot 2020-03-16 at 12 31 36
We're just using this provided API, which expects a string expressed in CSS units.

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
rbordeanu: Not possible I'm afraid this needs to be applied globally or it won't work :|

libs/angular-components/filter/filter-group/src/filter-group.component.html
rbordeanu: Actually... can't :(
Because of this:
Screenshot 2020-03-16 at 12 46 25

Material forces a shadow border unless a `class` elevation is present...

libs/angular-components/filter/filter-tags/src/_filter-tags.theme.scss
```diff
@@ -1,26 +1,39 @@
...
     }
   }
 
-  .selected-option-highlight,
-  .selected-option-highlight.mat-selected {
-    background-color: $filter-tags-background-color !important;
+  .mat-autocomplete-panel {
+    .mat-option {
+      // Todo: create dense version of mat-option
+      height: $filter-tags-option-height;
```
ttalbot: We can get rid of that by creating a dense version for autocomplete options.

PR #165, opened 2020-03-18
feat(input): implement fill appearance for input

beaolivei:
The main thing was to adjust the space between elements: icon prefix, input and icon suffix in the mat-form. 
There are many styles and possibilities for this component, so I tried to find a way that would look good in all cases. 
I make one special remark on the input type="number" here. It has in the original view a placeholder saying Amount that is not aligned. I am suggesting here that instead of this layout we share the HTML tag with an initial value of 0. 

💻 https://deploy-preview-165--vigorous-joliot-c90dab.netlify.com/ui-elements


themes/angular-theme/lib/form-field/_form-field-base.scss
```diff
@@ -3,3 +3,12 @@
     align-items: flex-end;
   }
 }
+
+label.mat-form-field-label {
+  left: 12px;
```
ttalbot: yes it's 12px when there's a mat-icon prefix but:
- it's 16px when there's no prefix
Screenshot 2020-03-19 at 17 53 21

- it's not indented when there's a text prefix
Screenshot 2020-03-19 at 17 53 35

themes/angular-theme/lib/form-field/_form-field-base.scss
```diff
@@ -3,3 +3,12 @@
...
+}
+
+.cdk-text-field-autofill-monitored:not(:-webkit-autofill) {
```
ttalbot: .cdk-stuff-autofill-monitored? why this over .mat-input-element?

PR #171, opened 2020-03-26
feat(global-nav): implem components and demo

Pi5rr5:
localhost_4200_global-nav
localhost_4200_global-nav (1)


libs/angular-components/global-nav/src/components/navbar/navbar.component.html
```diff
@@ -0,0 +1,17 @@
...
+      class="uxg-nav-header-button"
+      mat-icon-button
+      mat-primary
+      color="primary"
+      (click)="onMenuClick()"
+      aria-label="Menu"
+    >
+      <mat-icon>menu</mat-icon>
+    </button>
+    <a class="uxg-logo" mat-flat-button (click)="brandAction.emit($event)" dense></a>
```
ttalbot, line 13: I would not use a mat-flat-button here as it crops the company logo and creates an un-elegant grey ripple around it. It will also save you some lines of CSS. We just need to add a `cursor: pointer` in case uxg-logo is using `<a>` tag. But this should come from the theme.

libs/angular-components/global-nav/src/components/sidenav/sidenav.component.scss
```diff
@@ -0,0 +1,26 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+
+  .uxg-nav-list-main {
+    flex: 1;
+  }
+
+  .uxg-nav-header {
+    border-bottom: 1px solid #ececec;
```
ttalbot, line 11: No color in component or it will not support theming. You can already visualise the issue by switching to dark theme in the demo app.

PR #172, opened 2020-03-27
feat(menu): new entity menu component

dorsaffrigui:
Implement entity-menu

demo


libs/angular-components/entity-menu/src/entity-menu.component.scss
```diff
@@ -0,0 +1,40 @@
...
+  
+      .solutions-title {
+        margin-left: 10px;
+      }
+      .solution-list {
+        display: flex;
+        flex-wrap: wrap;
+  
+        .solution-item {
+          margin: 10px;
```
ttalbot, line 16: If you import spacing on top of your document, then here is the perfect place to use $uxg-spacing-2 for example.

libs/angular-components/entity-menu/src/entity-menu.component.scss
```diff
@@ -0,0 +1,40 @@
...
+          justify-content: center;
+          cursor: pointer;
+  
+          .solution-name {
+            text-align: center;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+            width: 100px;
+            margin: 8px 0
```
ttalbot, line 29: 8px = $uxg-spacing-2

PR #173, opened 2020-03-30
feat(banner): new banner component

Ping-YUAN:
Capture d’écran 2020-04-10 à 13 59 25


libs/angular-components/banner/src/_banner.theme.scss
```diff
@@ -0,0 +1,21 @@
+@mixin uxg-banner-theme($theme) {
+    $primary: map-get($theme, primary);
+    $accent: map-get($theme, accent);
+    $text-primary: mat-color($primary, default);
+    $color-accent: mat-color($accent, default);
+    $foreground: map-get($theme, foreground);
+    $text-banner: mat-color($foreground, base);
+
+    .uxg-banner-container{
+        .uxg-boday-1{
```
ttalbot: Typo here -> `.uxg-body-1`

libs/angular-components/banner/src/banner.component.scss
```diff
@@ -0,0 +1,29 @@
...
+    display: flex;
+    flex-direction: row;
+    .spliter {
+      margin-left: $uxg-spacing-2;
+      margin-right: $uxg-spacing-2;
+      height: 70px;
+      width: 1px;
+    }
+    .uxg-display-3 {
+      font-weight: bold;
```
ttalbot: Typography changes are expected in a dedicated typography mixin in the theme file.
If you need Bold font are you sure it won't be better to use headline scale?

Ping-YUAN:
🤔 from which browser you can't see the divider?

Ping-YUAN:
kind of strange, i use the same version, but can't reproduce it.

aifrim:
> kind of strange, i use the same version, but can't reproduce it.

happens to me as well
Chrome Version 80.0.3987.149 (Official Build) (64-bit)

Ping-YUAN:
i can reproduce it about no tooltip, but divider can't be reproduced

libs/angular-components/banner/src/banner.component.ts
```diff
@@ -0,0 +1,72 @@
...
+      if (banner.suffix) {
+        bannerText += ' ' + banner.suffix;
+      }
+      return bannerText;
+    }
+    return banner.value;
+  }
+  //prevent ui update to avoid tooltip doesn't work due to tooltip fresh
+  trackByFn = (index: number, item: BannerItem) => item.name;
+  ngOnInit() {}
```
aifrim, line 70: If you do not need these lifecycle hooks, remove them 😸

libs/angular-components/banner/src/model.ts
```diff
@@ -0,0 +1,7 @@
+export interface Banner {
```
aifrim, line 1: You have this and also the `BannerItem` present in `banner.component.ts:4`
I sugest you rename this to `BannerItem` and remove the one from the component file.

Ping-YUAN: the banner item is used inside banner component  they are different so i would like to keep them

code2nguyen:
@ttalbot the issue is fixed

new option for  small banner

Ping-YUAN:
> Please add a README.md at the root of the component ;)

readme doc added already please approve

PR #183, opened 2020-04-13
feat(wizard): improve wizard

rbordeanu:
- layout improvements
- add option to disable page titles
- active step coloring
- dark mode support:
Screenshot 2020-04-13 at 16 48 32


libs/angular-components/wizard/src/_wizard.theme.scss
```diff
@@ -5,14 +5,17 @@
...
+  $default-background-color: mat-color($background, card);
+
   $wizard-background-color: mat-color($background, background);
   $wizard-step-active-color: mat-color($accent, default);
   $wizard-step-done-color: map-get($uxg-grass, 100);
-  $wizard-timeline-color: map_get($mat-grey, 300);
+  $wizard-timeline-color: mat-color($background, status-bar);
 
   $wizard-dark-blue: #060546;
   $wizard-step-description-color: map-get($uxg-violet, 100);
```
ttalbot, line 16: Could it be replaced by `mat-color($primary, lighter);` ?

rbordeanu: Yes I was thinking of extending the palette but I think that's work to be done another PR.

PR #188, opened 2020-04-16
bugfix(expandable-table): fix header alignment

mnarcisa:


apps/angular-test-app/src/app/components/expandable-table-demo/expandable-table-demo.component.ts
```diff
@@ -23,12 +23,11 @@ export const COLUMNS = [
 export const GROUP_BY_KEY = 'apiGroup';
 
 @Component({
-  selector: 'uxg-expandable-table-demo',
+  selector: 'ffdc-expandable-table-demo',
```
aifrim, line 26: Should the selector be `uxg` like all the other components?

rbordeanu: this is the demo component, not the actual component

PR #189, opened 2020-04-20
fix(config): import should be the same for css and sass

ttalbot:
According to FDS documentation, import should look like this
```SCSS
@import '~@ffdc/uxg-angular-theme/all-theme';
```
Which is fine for an Angular SASS project but not compatible with a project running `ng new` choosing CSS for styling and trying to import `_all-theme.css`

### DIFF
--- a/themes/angular-theme/package.json
+++ b/themes/angular-theme/package.json
@@ -4,7 +4,7 @@
   "description": "Angular Material implementation of the Finastra Design System",
   "license": "MIT",
   "sass": "_all-theme.scss",
-  "css": "_all-theme.css",
+  "css": "all-theme.css",
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


code2nguyen:
we should not publish `_all-theme.scss`, 
In Admin-Portal project, developers still import this file in each component.scss

aifrim:
> we must not publish `_all-theme.scss`,
> In Admin-Portal project, developers still import this file in each component.scss
> 
> So in FDS documentation, import will be:
> `@import '~@ffdc/uxg-angular-theme/all-theme.css;`

If you check DevPortal you will see that we do not import this file everywhere.
But, my opinion, is that we still should publish this file, because developers may want to tweek some variables and build the theme.

Maybe some theme rework is required as to what files and structure we should have.

PR #194, opened 2020-04-29
Account card component

bcldvd:
## New *Account Card* component !

<br> 

image

Multiple things to consider here :
- [Skeleton equivalent](https://github.com/fusionfabric/finastra-design-system/blob/feature/account-card/libs/angular-components/cards/account-card/src/account-card.component.ts#L20-L26) of this component is provided in the same module  and re-uses the same styling
- [uxg-icon-category](https://github.com/fusionfabric/finastra-design-system/blob/feature/account-card/themes/angular-theme/lib/icon/_icon-theme.scss#L21-L27) class created in `angular-theme`
- Should the card have a fixed width or min-width is enough ?


code2nguyen:
Dont add nb-to-array and icon-caterogy into core, just create two separate lib..., 

`nb-to-array` should be `range pipe`....

and `icon-category` maybe should be only css in our theme.

PR #214, opened 2020-05-13
fix(table): fix accessibility issue for selected state

ttalbot:
It required a change in the palette to comply with the design (700 hue is a blue-ish violet 🤓) so I took the opportunity to also add dark-blue used in wizard as darkest hue of uxg-violet (900). It's still not compatible with theming (not part of $theme object) but it's a first step for a more comprehensive approach on our colors.

image


dxy009:
Hi :)
I share this color palette with you to show all the codes of colors in Sketch file. Had we changed totally the colors recently ?^^ 
Screenshot 2020-05-14 at 11 08 40
Best regards,
Xinyi

PR #217, opened 2020-05-18
fix(filter): improve filter tags

rbordeanu:
- allow full text search within the tag
- don't add the chip if not in the list
- highlight search term

### DIFF
--- a/libs/angular-components/filter/filter-tags/src/highlight.pipe.ts
+++ b/libs/angular-components/filter/filter-tags/src/highlight.pipe.ts
@@ -1,0 +1,17 @@
+import { PipeTransform, Pipe } from '@angular/core';
+
+@Pipe({ name: 'highlight' })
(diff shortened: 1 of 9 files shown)


libs/angular-components/filter/filter-tags/src/highlight.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { PipeTransform, Pipe } from '@angular/core';
```
aifrim, line 1: Should we move this to `core` and make it highlight everything if no `search` is present?

rbordeanu: This is pretty specific to the autocomplete search, I would keep it here.
And what do you mean highlight everything?

aifrim: As in the entire text.

PR #222, opened 2020-05-29
feat(entity-menu): Send full item in event emitter

bcldvd:


aifrim:
> Shouldn't the `IOnItemClick` be an item from the `data` array?
> 
> I suggest you create a `type` or an `interface` for the `data` input. Call it `EntityMenuItem` and then the `itemClick` can be like so:
> 
> ```ts
> @Input() data: Array<EntityMenuItem> = [];
> @Output() itemClick = new EventEmitter<EntityMenuItem>();
> 
> interface EntityMenuItem {
>   name: string;
>   abbr?: string;
> }
> ```
> 
> I am thinking that the `item` property is useless and redundant.
> And the abbreviation should only be added _(or not added at all)_ if not present in the first place because you can get
> 
> ```
> Republic of Singapore -> RO :(
> ```
> 
> Plus, can you rename `entity-menu.interfaces` to `entity-menu.models` so that we have the same naming convention everywhere?
> 
> And what is with the `@Input() property`? This is a presentation component. It should receive an `@Input() data: EntityMenuItem[]` and it should emit `EntityMenuItem`.
> 
> And then you can have something like this:
> 
> ```
> <mat-select
>   (selectionChange)="updateColumnMatcher($event.value)"
>   [(value)]="columnsMatcherExample"
> >
>   <mat-option *ngFor="let column of columnsExample" [value]="column.title">
>     {{ column.title }}
>   </mat-option>
> </mat-select>
> 
> <uxg-entity-menu
>   [title]="title"
>   [data]="dataSourceExample[columnsMatcherExample]"
>   [bottomLabel]="bottomLabel"
> >
> </uxg-entity-menu>
> ```
> 
> I do not get the need for the `bottomLabel`.
> 
> > the label of the pararaph in the bottom of the menu
> 
> If you want to `View All` you'd expect not to be seeing them all in the first place?
> Maybe add a search?
> 
> Sorry for the long comment. But I really think that this component needs a change 😸

As discussed, I will create an issue with improving this component and I will approve this PR.

PR #223, opened 2020-06-08
feat(config): add cypress image snapshot

ttalbot:
Clean PR for introducing cypress image snapshot on theme foundations and ui-elements.


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,31 @@
+name: Cypress tests
```
aifrim, line 1: Can you change the name to `Cypress Theme Tests` and trigger it also on PR, not only on push?

PR #224, opened 2020-06-10
fix(button-toggle): increase icon/text - background contrast

ttalbot:
Small contrast increase for content/background to pass Lighthouse and AXE accessibility tests (only for light theme)

Screenshot 2020-06-10 at 09 36 02


apps/angular-test-app/src/app/components/filter-panel-demo/filter-panel-demo.sample-data.ts
```diff
@@ -81,17 +81,17 @@ export const sampleFilterTags: Tag[] = [
 ];
 export const sampleToggleBtn: ToggleBtn[] = [
   {
-    label: 'API',
+    label: ' API',
```
aifrim, line 84: People should not be forced to prepend spaces to labels.

I suggest that the spacing is part of the toggle when a icon is present. What do you think?

ttalbot: I agree. @rbordeanu what do you think about it? I remember we had quite similar discussion about [input prefix here](https://github.com/fusionfabric/finastra-design-system/pull/165#pullrequestreview-384877329). Is it different in this situation?

rbordeanu: It's the same from my point of view, I still think we shouldn't enforce a spacing either way (on the theme level).
Because the code is just html bricks.
On the other hand, we could do it on the filter toggle *component* level, where we encapsulate the template on ours side.

rbordeanu: But here, for demo purposes you could just add a &nbsp; in the template directly I guess.

ttalbot: I removed the spaces so that it does not interfere with the real purpose of this PR, which is solving accessibility issues.

PR #226, opened 2020-06-24
update angular 9.1 and nx

code2nguyen:


aifrim:
As discussed. We will be upgrading directly to Angular 10.

PR #232, opened 2020-07-07
feat(entity-menu): Enable auxclick, ctrl and command click

bcldvd:


libs/angular-components/entity-menu/src/entity-menu.component.ts
```diff
@@ -9,6 +9,11 @@ import {
   OnChanges
 } from '@angular/core';
 
+export interface EntityMenuItemWEvent {
```
ttalbot, line 12: is it EntityMenuItem**W**Event on purpose or is it a typo?

bcldvd: on purpose "with event" :p 
For lack of a better name, any suggestions welcomed ;)

libs/angular-components/entity-menu/src/entity-menu.component.scss
```diff
@@ -1,48 +1,41 @@
...
 
 .container-list {
-    max-width: 427px !important;
-    
-    .mat-menu-content:not(:empty) {
-      padding: 0;
+  max-width: 427px !important;
+
+  .mat-menu-content:not(:empty) {
+    padding: 0;
```
ttalbot, line 7:
```suggestion
    padding: $uxg-spacing-0;
```

PR #238, opened 2020-07-24
feat(breadcrumb): add component

code2nguyen:
Capture d’écran 2020-07-24 à 16 38 07

### DIFF
--- a/apps/angular-test-app/src/app/components/breadcrumb-demo/breadcrumb-demo.component.html
+++ b/apps/angular-test-app/src/app/components/breadcrumb-demo/breadcrumb-demo.component.html
@@ -1,0 +1,59 @@
+<div class="wrapper">
+  <div class="title">
+    Default with menu
+  </div>
+  <mat-toolbar>
+    <uxg-breadcrumb [breadcrumbs]="defaultBreadcrumbs"></uxg-breadcrumb>
+  </mat-toolbar>
+
+  <div class="title">
+    Default with menu + primary color
+  </div>
+  <mat-toolbar>
+    <uxg-breadcrumb [breadcrumbs]="defaultBreadcrumbs" color="primary"></uxg-breadcrumb>
+  </mat-toolbar>
+
+  <div class="title">
+    Default without menu
+  </div>
+  <mat-toolbar>
+    <uxg-breadcrumb [breadcrumbs]="defaultBreadcrumbsWithoutMenu"></uxg-breadcrumb>
+  </mat-toolbar>
+
+  <div class="title">
+    Only one item
+  </div>
+  <mat-toolbar>
+    <uxg-breadcrumb [breadcrumbs]="onlyOneItem"></uxg-breadcrumb>
+  </mat-toolbar>
+
+  <div class="title">
+    Dense mode
+  </div>
+  <mat-toolbar>
+    <uxg-breadcrumb dense [breadcrumbs]="defaultBreadcrumbsWithoutMenu"></uxg-breadcrumb>
+  </mat-toolbar>
+
+  <div class="title">
+    Dense mode + accent color
+  </div>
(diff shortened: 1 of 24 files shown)


apps/angular-test-app/src/app/components/breadcrumb-demo/breadcrumb-demo.component.html
```diff
@@ -0,0 +1,59 @@
+<div class="wrapper">
```
aifrim, line 1: Can you please add some bottom spacing to this page?

libs/angular-components/breadcrumb/src/breadcrumb.component.html
```diff
@@ -0,0 +1,62 @@
+<ul class="uxg-breadcrumb" [attr.dense]="dense === '' ? '' : null" [attr.color]="color !== '' ? color : null">
+  <!-- back button -->
+  <ng-container *ngIf="shouldDisplayBackButton && enableBackButton">
+    <li class="uxg-breadcrumb-back">
```
ttalbot, line 4: What about using a mat-icon-button here ?

libs/angular-components/breadcrumb/src/breadcrumb.component.html
```diff
@@ -0,0 +1,62 @@
...
+          [ngTemplateOutletContext]="{ item: breadcrumbItem, isFirst: breadcrumbs.length == 1 }"
+        ></ng-template>
+      </ng-container>
+    </ng-template>
+  </ng-container>
+</ul>
+
+<!-- Item template -->
+<ng-template #linkItem let-breadcrumbItem="item" let-isFirstItem="isFirst">
+  <li [ngClass]="{ 'uxg-h6': isFirstItem }">
```
ttalbot, line 52: Can we make uxg-h6 optional, it's a behaviour we want for Application name in global-nav pattern, but not necessarily always on first item.

libs/angular-components/breadcrumb/src/breadcrumb.component.html
code2nguyen: There are some margins with mat-icon-button

libs/angular-components/breadcrumb/src/breadcrumb.component.html
code2nguyen: yes, do you have some suggestion for this optional feature ?

libs/angular-components/breadcrumb/src/breadcrumb.component.html
ttalbot: this kind of margin?
Screenshot 2020-07-28 at 10 33 19
that's acceptable 👍

libs/angular-components/breadcrumb/src/breadcrumb.component.html
ttalbot: Could be cool to allow users to set class using data input
```
{
   label: 'Design System',
   url: '/',
   class: 'uxg-h6'
 }
```
What do you think ?

code2nguyen: ok, I will add itemClass attribute in data model (to avoid class keyword)

libs/angular-components/breadcrumb/src/breadcrumb.component.html
code2nguyen: I tried, but I have some issue with alignement, so that keeps mat-icon for this moment.

apps/angular-test-app/src/app/components/breadcrumb-demo/breadcrumb-demo.component.html
code2nguyen: added

Issue #241, opened 2020-07-28
[Component schematic] add component-demo in angular-test-app


ttalbot:
**Is your feature request related to a problem? Please describe.**

When creating a component using our schematic
```sh
> npm run add:component
```
We still have to create manually the related demo component to the angular-test-app project

**Describe the solution you'd like**

I'd like to run
```sh
> npm run add:component my-component
> npm run start
```
accessing http://localhost:4200 and see the test app running with a page demonstrating my component.

**Describe alternatives you've considered**

Running
```sh
> ng generate component my-component-demo -m app.module.ts --style=scss
```
It saves some time but it's far for the ideal solution:
- my-component-demo is not created in components folder
- it does not include my-component
- it does not add routing to my-component-demo

**Additional context**

Does it make sense to add it in the angular-component schematic?
or should it be another schematic and execute them one after each other in the add:component script?

PR #248, opened 2020-08-04
feat(migration): migrate to angular 10

aifrim:
The `component` schematic is now broken.
I would like to have it fixed within this issue https://github.com/fusionfabric/finastra-design-system/issues/241 because they basically cover the same fix, but in different parts.

DIsabled PR CI Linting due to:
```
fatal: path 'tsconfig.base.json' exists on disk, but not in 'remotes/origin/master'
```

Will make PR to update it after this reaches master.


code2nguyen:
We must update peerDeps in https://github.com/fusionfabric/finastra-design-system/blob/feature/angular-10-migration/libs/angular-components/package.json

code2nguyen:
can we use the lastest version of angular 10.0.7

code2nguyen:
dont extend tsconfig.ts in tsconfig.app.ts ...example: https://github.com/fusionfabric/finastra-design-system/blob/feature/angular-10-migration/libs/angular-components/popover/tsconfig.lib.json

aifrim:
> We must update peerDeps in https://github.com/fusionfabric/finastra-design-system/blob/feature/angular-10-migration/libs/angular-components/package.json

This was already part of the repo before hand. And it is the pattern we use.

aifrim:
> can we use the lastest version of angular 10.0.7

Best I can do it 10.0.4 (@nrwl/nx@10.0.7)

PR #251, opened 2020-08-14
initial commit for the toaster and toaster demo component

achilehero:
Moved the toaster component used in DevPortal and UXP into FDS


libs/angular-components/toaster/src/_toaster.theme.scss
```diff
@@ -0,0 +1,59 @@
...
+    box-shadow: 0 6px 10px 0 rgba(0, 0, 0, 0.07), 0 1px 18px 0 rgba(0, 0, 0, 0.06), 0 3px 5px -1px rgba(0, 0, 0, 0.1);
+    .icon-container {
+      width: $container-size;
+      border-radius: $border-radius 0 0 $border-radius;
+      padding: $uxg-spacing-3;
+      mat-icon {
+        color: white;
+      }
+    }
+    .message {
```
adrian-apetrei, line 26: Text color is not visible in dark mode 
image

libs/angular-components/toaster/src/toaster.component.html
```diff
@@ -0,0 +1,11 @@
+<div fxLayoutAlign="center center" class="icon-container">
+  <mat-icon dense *ngIf="toast.config.type === 'success'">check_circle_outline</mat-icon>
+  <mat-icon dense *ngIf="toast.config.type === 'error'">error_outline</mat-icon>
+  <mat-icon class="material-icons-outlined" dense *ngIf="toast.config.type === 'warning'">report_problem</mat-icon>
+  <mat-icon class="material-icons-outlined" dense *ngIf="toast.config.type === 'info'">info</mat-icon>
+  <!-- <mat-icon dense *ngIf="toast.config.type === 'warning'" svgIcon="warning"></mat-icon>
```
adrian-apetrei, line 6: remove commented code

apps/angular-test-app/src/app/components/toaster-demo/toaster-demo.component.scss
```diff
@@ -0,0 +1,35 @@
+$uxg-toolbar-height: 56px;
+
+uxg-toaster-container {
```
adrian-apetrei, line 3: Dark mode toggle is not accessible when toaster is open

adrian-apetrei (changes requested):

libs/angular-components/toaster/src/toaster.component.html
achilehero: Removed

libs/angular-components/toaster/src/_toaster.theme.scss
achilehero: image
fixed

apps/angular-test-app/src/app/components/toaster-demo/toaster-demo.component.scss
achilehero: Fixed

adrian-apetrei (approved):

libs/angular-components/toaster/src/_toaster.theme.scss
```diff
@@ -0,0 +1,63 @@
...
+    border-radius: $border-radius;
+    color: $toaster-text-color;
+    background: mat-color($background, card);
+    box-shadow: 0 6px 10px 0 rgba(0, 0, 0, 0.07), 0 1px 18px 0 rgba(0, 0, 0, 0.06), 0 3px 5px -1px rgba(0, 0, 0, 0.1);
+    .icon-container {
+      width: $container-size;
+      border-radius: $border-radius 0 0 $border-radius;
+      padding: $uxg-spacing-3;
+      mat-icon {
+        color: white;
```
ttalbot, line 27: if we want icon-container colors to be themable then we cannot assume white icon will fit everytime. So I encourage the use of [contrast associated with container color](https://design.fusionfabric.cloud/foundations/colors?tab=dev#typography) instead.

libs/angular-components/toaster/src/_toaster.theme.scss
```diff
@@ -0,0 +1,63 @@
...
+    .message {
+      padding: $uxg-spacing-3;
+      word-break: break-word;
+      white-space: pre-line;
+    }
+    &.destroy-by-click {
+      cursor: pointer;
+    }
+    &.type-success .icon-container {
+      background-color: map-get($uxg-grass, 100);
```
ttalbot, line 39:
```suggestion
      background-color: mat-color($uxg-grass, 100);
      color: mat-contrast($uxg-grass, 100);
```
and so on for the others

ttalbot (changes requested):

achilehero:
> 
> 
> Looks like the toast is moving diagonally when entering and leaving, is it intended transition?
> 
> toast_diagonal_transition

It's because the top margin is also animated, and it goes from 0 to the actual value. I can remove that if you think it's nicer to just be an horizontal slide. It's an easy one. Just let me know

achilehero:
Sure. I'll make it horizontal!

libs/angular-components/toaster/src/_toaster.theme.scss
```diff
@@ -0,0 +1,63 @@
+@import 'angular-theme/lib/core/style/spacing';
+@import 'angular-theme/theme-variables';
+@import 'angular-theme/lib/core/layout/breakpoints';
+
+@mixin uxg-toaster-theme($theme) {
```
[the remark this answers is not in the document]
achilehero: I agree with the separation, but it seemed to me that in this particular case, since apart from the colors and text there are only a few spacing and sizing things present here, it wouldn't be worth splitting them into separate files. What do you think?

libs/angular-components/toaster/src/_toaster.theme.scss
achilehero: If I use something like:
`&.type-warning .icon-container {
   background-color: mat-color($uxg-gold, 100);
      mat-icon {
        color: map-get($uxg-gold, default-contrast);
      }
    }`
I would get all the icons black, for all the backgrounds:
image

Is this what we want?

rbordeanu: I don't really like this... :(  @ttalbot what other options do we have?

ttalbot: Well, if we target AA we have 2 options to increase contrast:
- darken the background (very unlikely, we already use 100% opacity secondary colors)
- darken the icon 😕

achilehero: In this case, I guess that making the icons area like [here](https://finastra.invisionapp.com/d/#/console/18537805/406609402/preview) would be better, then? I understand accessibility, but I don't like either how the icons look black on colored background.

ttalbot: In this case we will need to implement the [darker versions of the secondary colors](https://finastra.invisionapp.com/d/#/console/18537805/385445443/preview) which is not really a problem.

achilehero: This is what we get after putting in the darker versions of secondary colors and using a lighter background:
image

libs/angular-components/toaster/src/_toaster.theme.scss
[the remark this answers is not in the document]
achilehero: Separated the files

libs/angular-components/toaster/src/_toaster.theme.scss
```diff
@@ -0,0 +1,63 @@
...
+  $background: map-get($theme, background);
+
+  uxg-toaster {
+    margin: $uxg-spacing-3 4vw;
+    min-height: $container-size;
+    width: 92vw;
+    border-radius: $border-radius;
+    color: $toaster-text-color;
+    background: mat-color($background, card);
+    box-shadow: 0 6px 10px 0 rgba(0, 0, 0, 0.07), 0 1px 18px 0 rgba(0, 0, 0, 0.06), 0 3px 5px -1px rgba(0, 0, 0, 0.1);
```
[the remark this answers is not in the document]
achilehero: I replaced the shadow with elevation

libs/angular-components/toaster/src/_toaster.theme.scss
achilehero: So, @ttalbot, are we OK with the changes? If so, maybe we can try merging it.

ttalbot (approved):

PR #252, opened 2020-08-24
feat(filter-group): improve horizontal filter click to open/close

adrian-apetrei:


rbordeanu (changes requested):
My suggestion is to add a `clickOutside` directive that could be reused in other situations, if needed.

aifrim (approved):

ttalbot (approved):

rbordeanu (approved):

PR #256, opened 2020-09-03
fix(theme): make dark theme AA compliant

Octavians98:


rbordeanu:
> Here is the before / after
> It's still AA according to Lighthouse

Ugh, I really don't like at all how black text looks on violet/fuchsia...
What other choice do we have?

Call me crazy, but to my naked eye white text color is more readable than black text in this situation.
How can AA consider this a better contrast? :S

achilehero:
I tend to agree with Razvan with black font on violet/fuchsia. We had the same issue on the notifications. To answer to you question, Razvan, on how AA seems better, I think it's because the computer analyzes only color vs background color for the button. But we, as humans, also perceive the black baground of the theme around the buttons, and that makes it look weird.

rayah:
We can use white text on the primary buttons and on fuschia (passes AA).

ttalbot (approved):

dorsaffrigui (approved):

rayah (approved):

bcldvd (approved):

rbordeanu (approved):

PR #271, opened 2020-10-02
feat(global-search): Recent searches

Octavians98:
This PR aim is to add search history to the global search.
Recent searches Global Search@2x


libs/angular-components/global-search/src/components/global-search-overlay/global-search-overlay.component.ts
```diff
@@ -66,6 +67,7 @@ export class GlobalSearchOverlayComponent implements AfterViewInit {
   @Output() searchTermChange: EventEmitter<string> = new EventEmitter<string>();
   @Output() itemClicked: EventEmitter<string> = new EventEmitter<string>();
   @ViewChild('searchInput', { static: true }) searchInput!: ElementRef;
+  @ViewChild('searchIcon', { static: true }) searchIcon!: MatIcon;
```
rbordeanu, line 70: Why not an ElementRef as well? To avoid the extra import.

adrian-apetrei (approved):

rbordeanu (approved):

PR #305, opened 2020-12-03
feat(config): use standard-version to generate changelogs

ttalbot:
### Description:
Replace previous release script to use [standard-version](https://github.com/conventional-changelog/standard-version) instead. ✌️ 
now run:
```BASH
npm run release
```

The benefits are:
- the process is less error prone
- the tool guess the new version number based on conventional commit messages. [Versions can be forced](https://github.com/conventional-changelog/standard-version#release-as-a-target-type-imperatively-npm-version-like) using `--release-as` param when needed
- CHANGELOG.md are generated separately for theme and components
- commit and tag are created
- release is done automatically
- publish should be triggered automatically, I hope 🤞 

### IMPROVEMENTS:
- [x] Create GitHub release in [release action](https://github.com/fusionfabric/finastra-design-system/blob/master/.github/workflows/release.yml)
- [ ] Include latest part of the changelog in release body


dorsaffrigui (approved):

rayah (approved):

package.json
```diff
@@ -38,7 +42,9 @@
     "help": "nx help",
     "nx": "nx",
     "testjest": "jest",
-    "release": "node ./scripts/release.js"
+    "release": "npx standard-version -an",
```
bcldvd, line 45:
```suggestion
    "release": "npx standard-version -an && git push --follow-tags",
```

bcldvd (approved):

rbordeanu (approved):

PR #311, opened 2020-12-24
fix(card): product card text color in dark theme

ArtemSilka:
Hey @ttalbot. Should be good to go. now when you switch to dark mode, card's text color turns white, and back to black when dark mode is off.


rbordeanu (changes requested):

rbordeanu:
Hey @ArtemSilka thanks for your contribution.
Unfortunately this does not solve the problem, see my comments and suggestion.

PR #347, opened 2021-04-16
feat(theme): use css var for background, apply to body

ttalbot:


dorsaffrigui (approved):

dorsaffrigui (changes requested):
The theme editor doesn't work with the dark theme

dorsaffrigui (approved):

bcldvd (approved):

PR #357, opened 2021-05-18
fix(filter-tags): Add focus and blur events

adelinachiripuci:


mnarcisa:
Can you keep the focus effect even when you're selecting an option, not only on the input? Otherwise, there will be a flicker on the label that it's quite visible and we want to prevent that.

image

apps/angular-test-app/src/app/components/filter-panel-demo/filter-panel-demo.component.scss
```diff
@@ -54,6 +54,10 @@
   padding: 0 25px 0 25px;
 }
 
+.focused {
+  color: blue;
```
dorsaffrigui, line 58: Can you please use another color extracted from the theme ( like primary or secondary) to have it accessible both on light and dark theme [
Screenshot 2021-05-19 at 16 45 16
](url)

ttalbot: or even not changing the style at all, let's keep real world use cases. For testing purpose you can use right container.

ttalbot (approved):

ttalbot (approved):

mnarcisa (approved):

adrian-apetrei (approved):

PR #402, opened 2021-08-24
fix(fds-button): gradient background

dorsaffrigui:
This PR solves the problem we had with the linear gradient which is not compatible with the background-color css property and therefore it wan't possible to apply a gradient color to mdc/mwc buttons.

A possible solution would be to use background property instead of background-color and here is the result : 

Screenshot 2021-08-25 at 13 50 11


rbordeanu:
Can you please attach a screenshot and a small description of what issue this PR solves?

rbordeanu:
Why are there so many formatting changes? 
Feels like the formatting of those files should have been correct in the first place.

ttalbot (approved):

PR #409, opened 2021-09-20
feat(tabs): new tabs

adrian-apetrei:
image


ttalbot (approved):

rbordeanu:
When clicking on the tabs to the right of the current one, the animations looks ok.
When clicking to the left, it's not how it's supposed to be (compared with regular mat tabs).

rbordeanu (changes requested):
When clicking on the tabs to the right of the current one, the animations looks ok.
When clicking to the left, it's not how it's supposed to be (compared with regular mat tabs).

rbordeanu (changes requested):
There is one more small thing:
When switching the tabs, the label shifts a couple of pixels between the active and the inactive states.

rbordeanu (approved):

PR #491, opened 2021-12-07
Feature/update lit 2

code2nguyen:


.storybook/fds-theme.js
```diff
@@ -0,0 +1 @@
+import { create } from '@storybook/theming';
```
ttalbot, line 1: Is it part of another improvement?

ttalbot (approved):

.storybook/fds-theme.js
code2nguyen: yes, removed

PR #494, opened 2021-12-07
Migration to Angular 13

achilehero:
Hi guys!
I started looking into upgrading the DS to Angular 13. As usual, there is no easy way with these things :)
In this case here, the big change is that @angular/material 13 has dropped tilde imports. So changes like in `libs/angular-components/video-thumbnail/src/video-thumbnail.component.scss` are working and are fine `~@angular/material` changed to `@angular/material`. 
However, we have two use cases, in https://github.com/Finastra/finastra-design-system/blob/30e4d8dc74a78ee5ab91a0d7d37e48fe37a1d0cd/themes/angular-theme/lib/core/_core.scss#L2 and https://github.com/Finastra/finastra-design-system/blob/30e4d8dc74a78ee5ab91a0d7d37e48fe37a1d0cd/themes/angular-theme/lib/core/theming/_palette.scss#L4 where it doesn't work, either with `~` or without it.

While I am further investigating on how to solve it, if anyone has any idea, please shoot!


ttalbot (approved):
Thank you for upgrading 🙏 

You removed ~ from @use statements on @angular/material, I think it's no longer necessary even for other packages.
This can be part of another PR.

angular.json
```diff
@@ -348,15 +341,9 @@
           "options": {
             "browserTarget": "fds-test-app:build"
           }
-        },
```
achilehero, line 351: To be checked

achilehero:
There is an open issue in the nx repo: https://github.com/nrwl/nx/issues/7844. I already tried the suggested changes in the `jest.preset.js` but it doesn't seem to be working, though. Some other things I found on the topic: https://github.com/thymikee/jest-preset-angular/issues/1090 and https://github.com/thymikee/jest-preset-angular/issues/1149 (which is actually pointing me to the first issue)

dorsaffrigui (approved):

blminami (approved):

PR #565, opened 2022-01-21
fix(fds-avatar): generate warnings instead of errors

dorsaffrigui:


libs/web-components/avatar/src/avatar.ts
```diff
@@ -27,18 +27,18 @@ export class Avatar extends LitElement {
 
   render() {
     if (this.primary && this.secondary) {
-      throw new Error('Cannot use both primary and secondary attribute');
+      console.warn('Cannot use both primary and secondary attribut');
```
ttalbot, line 30:
```suggestion
      console.warn('Cannot use both primary and secondary attribute');
```

libs/web-components/avatar/src/avatar.ts
```diff
@@ -27,18 +27,18 @@ export class Avatar extends LitElement {
...
     }
-    if (this.dense && this.large) {
-      throw new Error('Cannot use both dense and large attribute');
+    if (this.large && this.dense) {
+      console.warn('Cannot use both large and dense attribut');
```
ttalbot, line 33:
```suggestion
      console.warn('Cannot use both large and dense attribute');
```

bcldvd (approved):

ttalbot (approved):

PR #574, opened 2022-01-28
fix(fds-avatar): add avatar size css vars

dorsaffrigui:


dorsaffrigui:
> Can we imagine having only one custom prop like `--fds-avatar-size` since width and height are equal (it's a perfect circle) ?



> Can we imagine having only one custom prop like `--fds-avatar-size` since width and height are equal (it's a perfect circle) ?

Yes i like this idea 👍🏼

FrancineO (approved):

bcldvd (approved):

ttalbot (approved):

PR #587, opened 2022-02-09
fix(user-profile): replace fds avatar with mwc icon button

dorsaffrigui:


libs/web-components/user-profile/src/styles.scss
```diff
@@ -4,11 +4,10 @@
   display: inline-block;
   --mdc-menu-min-width: 300px;
 
-  &>fds-avatar {
-    &:hover {
-      cursor: pointer;
-    }
-  }
+   mwc-icon-button  {
+    display: block;
```
Pi5rr5, line 8: you should remove the `display: block` to fix this:
image

libs/web-components/user-profile/src/styles.scss
```diff
@@ -4,11 +4,10 @@
...
+    --mdc-ripple-color: var(--fds-primary);
```
Pi5rr5, line 9: does the design system guideline have a color on the `--mdc-ripple-color` ?

libs/web-components/user-profile/src/styles.scss
dorsaffrigui: Screenshot 2022-02-10 at 13 24 11

libs/web-components/user-profile/src/styles.scss
dorsaffrigui: We have those guidelines, it's similar to material Screenshot 2022-02-10 at 13 32 17

FrancineO (approved):

ttalbot (approved):

PR #602, opened 2022-02-18
feat(color): prefer error color over warn

ttalbot:


themes/angular-theme/lib/badge/_badge-theme.scss
```diff
@@ -28,7 +28,9 @@
 
     @include uxg-badge-color('primary', $primary);
     @include uxg-badge-color('accent', $secondary);
-    @include uxg-badge-color('warn', $warn);
+    @include uxg-badge-color('secondary', $secondary);
+    @include uxg-badge-color('error', $error);
+    @include uxg-badge-color('warn', $error);
```
dorsaffrigui, line 33: Do we need to include the warn color here ?

dorsaffrigui (approved):

themes/angular-theme/lib/badge/_badge-theme.scss
ttalbot: we keep the warn option for compatibility purpose, we will have to find a strategy for marking it deprecated and then removing it.

BeataKr (approved):

achilehero (approved):

Issue #632, opened 2022-03-09
Radio


Pi5rr5:
## Specs

![Screenshot 2022-03-09 at 13 46 17](https://user-images.githubusercontent.com/20640264/157444619-9d620e6e-6b19-4baf-a036-1998771aa716.png)
https://www.figma.com/file/E1Mb1556RT3HbAUVu2Q0LV/Finastra-design-system?node-id=48606%3A19619

## Tasks

- [x] Generate component
- [x] Demo
- [x] Storybook
- [x] Tests
- [x] Doc
- [ ] Release/Publish

PR #638, opened 2022-03-14
feature(#632): fds-radio

Pi5rr5:
Radio #632


libs/web-components/radio/README.md
```diff
@@ -0,0 +1,14 @@
+# Radio
+
+[![See it on NPM!](https://img.shields.io/npm/v/@finastra/radio?style=for-the-badge)](https://www.npmjs.com/package/@finastra/radio)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@finastra/radio?style=for-the-badge)](https://bundlephobia.com/result?p=@finastra/radio')
+
```
bcldvd, line 5: Missing storybook link, like this one: 
[Storybook](https://finastra.github.io/finastra-design-system/?path=/story/components-app-bar--default)

libs/web-components/radio/src/radio.ts
```diff
@@ -0,0 +1,21 @@
+import { RadioBase } from '@material/mwc-radio/mwc-radio-base';
+import { customElement } from 'lit/decorators.js';
+import { styles } from './styles.css';
+
+/**
+ * @cssprop {color} [--mdc-theme-secondary=#917EE0] - FDS Primary color
```
bcldvd, line 6: Would have rather set `--fds-primary` since it'll be the public API.
The mapping we do to material could change if we decide to move away from it

libs/web-components/radio/test/radio.test.ts
```diff
@@ -0,0 +1,11 @@
+import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
+import '../src/radio.js';
+import { Radio } from '../src/radio.js';
+
+describe('Radio', () => {
+  it('loads accessibly', async () => {
+    const el: Radio = await fixture(html`<fds-radio aria-label="test"></fds-radio>`);
```
bcldvd, line 7:
```suggestion
    const el: Radio = await fixture(html`<fds-radio></fds-radio>`);
```

libs/web-components/radio/stories/radio.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+    reducedTouchTarget: {
+      table: {
+        category: 'attributes',
+        defaultValue: { summary: false }
+      },
+      control: 'boolean',
+      defaultValue: false,
+      type: 'boolean',
+      description: 'When true, the radio removes touch target that extends beyond visual boundary of the component.'
+    },
```
bcldvd, line 71: Missing some properties like `aria-label` (https://github.com/material-components/material-web/blob/master/packages/radio/mwc-radio-base.ts#L134)

libs/web-components/radio/stories/radio.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+  },
```
bcldvd, line 72: Should we document all callbacks ? https://github.com/material-components/material-web/blob/master/packages/radio/mwc-radio-base.ts#L258-L267

libs/web-components/radio/stories/radio.stories.ts
Pi5rr5: `aria-label` is native api like `label` or `aria-labelledby` so I don't think it's relevant to document it 😉

libs/web-components/radio/stories/radio.stories.ts
Pi5rr5: As mention in the `README` we extends mwc radio base, so a link to their documentation should be enough 🤓
https://github.com/Finastra/finastra-design-system/blob/feature/fds-radio/libs/web-components/radio/README.md#example-usage

Pi5rr5:
> black on black unselected state with dark mode image

Fixed 👍

ttalbot (approved):

dorsaffrigui (approved):

PR #675, opened 2022-04-05
aria-labels on table

BeataKr:
small, but still something new ;)

### DIFF
--- a/libs/angular-components/table/src/table.component.html
+++ b/libs/angular-components/table/src/table.component.html
@@ -1,79 +1,49 @@
 <div class="uxg-data-table">
-  <table
-    mat-table
-    matSort
-    cdkDropListGroup
-    [dataSource]="dataToComponent"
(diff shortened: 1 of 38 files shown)


libs/angular-components/table/src/table.component.html
```diff
@@ -124,25 +122,33 @@
...
           mat-icon-button
+          [attr.aria-label]="'Edit ' + ariaRowString(row)"
           (click)="rowEditClick(row); $event.stopPropagation()"
         >
           <mat-icon class="material-icons-outlined">edit_outline</mat-icon>
         </button>
         <button
           *ngIf="!row.uxgTableEdit && enableDelete"
           mat-icon-button
+          [attr.aria-label]="'Delete ' + ariaRowString(row)"
```
BeataKr, line 133: still trying to find a better way for that

PR #700, opened 2022-04-21
Feature/changelog updates

adelinachiripuci:


blminami:
We can play a bit more with the contrast for `bugfix` label on dark mode - maybe something similar as for `update`? (no red background)

image

blminami:
+++ TBD:
I would apply the same opacity to all labels on dark mode, as here:

image

vs.

image

- better accessibility / contrast score
- we have a pattern for these labels (not only update and bugfix with lower opacity)

dorsaffrigui (approved):

blminami:
I think it would be better to declare a new interface for categories, smth like

export interface Category {
   label: string;
   link?: string;
}
and as part of the Release model, categories: Category[]

html changes:

<a href="{{ category.link }}" rel="noreferrer" target="_blank" class="mat-button-base">
{{ category.label }}
</a>
Also, not sure if the link prop should be mandatory (so add only for the categories that provide a link).
We should allow categories without links, what do you think? It might be the case for DP to have links, but not sure about other services :)

blminami (approved):

bcldvd (approved):

PR #721, opened 2022-05-13
disabled step style

dorsaffrigui:


libs/web-components/stepper/src/v-styles.scss
```diff
@@ -83,6 +82,13 @@
     align-self: center;
     padding: var(--fds-spacing-3, 16px);
   }
+
+  .disabled {
```
ttalbot, line 86: Is it necessary? it's already defined in base-style.scss

ttalbot (approved):

PR #743, opened 2022-06-13
feat: add categories & new storybook style

Pi5rr5:


libs/web-components/chip/stories/chip.stories.ts
```diff
@@ -1,12 +1,12 @@
...
-import { html } from 'lit-html';
 import '@finastra/chip';
 import type { Chip } from '@finastra/chip';
+import { Meta, Story } from '@storybook/web-components';
+import { html } from 'lit-html';
 import { argTypes, cssprops } from './custom-element.json';
 
 export default {
-  title: 'Components/Chip',
+  title: 'DATA DISPLAY/Chip',
```
dorsaffrigui, line 9: A chip could be an action as well

Pi5rr5: Yup but the purpose of the chips is to display data, not like a button for example that's why ! 🤓

ttalbot (approved):

bcldvd (approved):

PR #750, opened 2022-06-15
Fab component added 

leakhairallah:


libs/web-components/fab/src/styles.scss
```diff
@@ -0,0 +1,97 @@
...
+        height: 20px;
+        top: 14px;
+        left: 48px;
+        letter-spacing: 1.25px;
+    }
+}
+
+  
+:host([dense]) {
+    .mdc-fab {
```
bcldvd, line 36: include only what is different in the variants (both dense and gradient)

code2nguyen:
> the branch name should be like feature/fds-fab and the package-lock file should not be committed

package-lock ?, pourquoi on commit pas, si on commit pas, le build ne marche pas ?

Ping-YUAN:
> > the branch name should be like feature/fds-fab and the package-lock file should not be committed
> 
> @Ping-YUAN package-lock ?, pourquoi on commit pas, si on commit pas, le build ne marche pas ?

if no changes in package.json. i didn't see any reason to commit package-lock.json

dorsaffrigui:
> > > the branch name should be like feature/fds-fab and the package-lock file should not be committed
> > 
> > 
> > @Ping-YUAN package-lock ?, pourquoi on commit pas, si on commit pas, le build ne marche pas ?
> 
> if no changes in package.json. i didn't see any reason to commit package-lock.json

I agree with Ping, we should revert the changes in the package-lock.json

libs/web-components/fab/src/styles.scss
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    .mdc-fab--extended{
+        min-width: 120px;
+        width: auto;
+    }
+
+    .mdc-fab__label{
+        @include fds.typography(body-3);
+        @include fds.property(color, on-surface);
```
ttalbot, line 24: If bg-color is primary then label should be on-primary.
Is it necessary since you already specified text-color on `.mdc-fab` ?

PR #812, opened 2022-07-29
feat(textfield): allow custom radius with CSS var

ttalbot:


libs/web-components/textfield/src/textfield.ts
```diff
@@ -9,6 +9,7 @@ import { styles } from './styles.css';
  * @cssprop {color} [--fds-primary=#694ED6] - Textfield color
  * @cssprop {color} [--fds-icon-color=#694ED6] - Icon color.
  * @cssprop {color} [--fds-icon-trailing-color=#694ED6] - Icon trailing color.
+ * @cssprop {string} [--fds-text-field-radius=4px] - Border radius of the outline.
```
ttalbot, line 12:
```suggestion
 * @cssprop {text} [--fds-text-field-radius=4px] - Border radius of the outline.
```

bcldvd (approved):

Ping-YUAN (approved):

libs/web-components/textfield/src/styles.scss
```diff
@@ -5,8 +5,6 @@
 $textfield-icon-color: var(--fds-primary, #694ed6);
 $textfield-icon-trailing-color: var(--fds-on-surface-medium, color.change(#000, $alpha: 0.54));
 
-
-
 :host {
   padding-top: 24px;
```
Ping-YUAN, line 9: I have one question here during I try to use textfield component. 

Typical, we can add a class to <fds-textfield> tag to add a padding if we need some padding. 
I didn't see the necessity to add it inside textfield component.

Issue #785, opened 2022-07-07
Pie/Donut chart stop working past 7 values


bcldvd:
**Describe the bug**
Both Pie and Donut charts work well with 7 values. Starting with 8 and above, they don't render

**Code sample to reproduce**
```html
<script type="module" src="node_modules/@finastra/charts/dist/src/pie-chart.js"></script>
<script type="module" src="node_modules/@finastra/charts/dist/src/donut-chart.js"></script>
<div class="main-content">
    <div>Default</div>

    <div class="card">
        <fds-pie-chart id="pieChart" width="380"></fds-pie-chart>
    </div>
</div>
<script>
    const pieCharts = document.querySelectorAll('#pieChart');
    pieCharts.forEach((pieChart) => {
        pieChart.data = [44, 55, 13, 43, 22, 12, 14, 28]
        pieChart.labels = ['Apple', 'Mango', 'Orange', 'Watermelon', 'Wiki', 'Strawberry', 'Banana', 'Lemon']
        
    })
</script>
<style>
    .card {
        background-color: #fff;
        padding: 40px;
        border-radius: 8px;
        display: flex;
        flex-direction: column;
        justify-content: center;
        align-items: center;
        height: 500px;
        width: 400px
    }
</style>
```

**Expected behavior**
Charts should render with any number of values

**Screenshots**
<img width="1206" alt="image" src="https://user-images.githubusercontent.com/569053/177738615-8e368495-89ac-4c19-8afe-d797f70c2e69.png">

PR #816, opened 2022-08-02
fix(wc): Pie/Donut chart stop working past 7 values

code2nguyen:
- issue #785

### DIFF
--- a/libs/web-components/charts/src/pie-chart.ts
+++ b/libs/web-components/charts/src/pie-chart.ts
@@ -30,8 +30,8 @@ export class PieChart extends ApexChartsWrapper {
         return this._data;
     }
     public set data(value: number[]) {
-        this._data = value.sort((a, b) => b - a);
-        this.series = value
+        this._data = value;
+        this.series = this._data;
     }
 
     private _labels: string[] = [];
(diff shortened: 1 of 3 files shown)


libs/web-components/charts/src/pie-chart.ts
```diff
@@ -30,8 +30,8 @@ export class PieChart extends ApexChartsWrapper {
         return this._data;
     }
     public set data(value: number[]) {
-        this._data = value.sort((a, b) => b - a);
```
bcldvd, line 33: Sorting is lost. Need to have it

code2nguyen: it create an error, because data and label need to be sync.

bcldvd: Can we have the data sorted some other ways ? :)

code2nguyen: This not link with this issue. Please create an issue about it if you need. Please considering about following points:

- The sync between data and label is more expensive some time can create errors if user change data -> sort data -> sort label ... then user update labels but list of labels does not sync back from chart component.
- Dont know which order (desc/asc) is better, it depends de lire direction (rtl or ltr) or event depends on order of theme colors.
-User normalement, can easily transform data outside the component. Component will display as its input. It make component more predicable.

dorsaffrigui (approved):

PR #830, opened 2022-08-09
feature: fds-icon

Pi5rr5:
image


libs/web-components/icon/src/styles.scss
```diff
@@ -0,0 +1,38 @@
+@use './_fds-icon';
+@use '@finastra/fds-theme' as fds;
+
+:host {
+  @extend %material-icons;
```
dorsaffrigui, line 5: The default icon color isn't from the fds-theme, please use the fds-on-background by default

libs/web-components/icon/src/styles.scss
```diff
@@ -0,0 +1,38 @@
...
+:host([dense]) {
+  --fds-icon-size: 18px;
+}
+
+:host([large]) {
+  --fds-icon-size: 36px;
+}
+
+:host([extra_large]) {
+  --fds-icon-size: 48px;
```
dorsaffrigui, line 17:
```suggestion
  --fds-icon-size: var(--fds-spacing-5, 48px);
```

Pi5rr5: spacing isn't for icon size, the day we have fds-icon-size variable it will make sense to use it

libs/web-components/icon/src/styles.scss
Pi5rr5: The icon color inherit from the context, if we set fds-on-background by default it will override this behaviour and we will have to update everywhere the icon is used to set the color.

bcldvd (approved):

dorsaffrigui (approved):

Issue #821, opened 2022-08-04
[textarea] apply same outlined style than textfield


ttalbot:
**Describe the bug**
fds-textarea use old underline style but should match textfield new outlined style.

**To Reproduce**

1. Go to [Storybook](https://finastra.github.io/finastra-design-system/?path=/story/forms-textarea--default)
2. See old style

**Expected behavior**
fds-textarea should adopt same style than textfield

**Screenshots**
<img width="324" alt="image" src="https://user-images.githubusercontent.com/371041/182800365-f44360f9-52ee-425a-ac07-6243d93ce766.png">

Issue #820, opened 2022-08-04
[select] apply same outlined style than textfield


ttalbot:
**Describe the bug**
fds-select use old underline style but should match textfield new outlined style.

**To Reproduce**

1. Go to [Storybook](https://finastra.github.io/finastra-design-system/?path=/story/forms-select--default)
2. See old style

**Expected behavior**
fds-select should adopt same style than textfield

**Screenshots**
<img width="289" alt="image" src="https://user-images.githubusercontent.com/371041/182795853-da84e407-97a1-4e12-9845-7873bda6d730.png">

PR #853, opened 2022-08-26
bugfix#820-821 use new textfield guideline

Pi5rr5:
#821 
image
#820 
image

### DIFF
--- a/libs/web-components/select/src/styles.scss
+++ b/libs/web-components/select/src/styles.scss
@@ -1,12 +1,75 @@
 @use "@material/select/mdc-select.scss";
-@use "@finastra/fds-theme"as fds;
+@use "@finastra/fds-theme" as fds;
 
 :host {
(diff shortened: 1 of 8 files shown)


libs/web-components/select/src/styles.scss
```diff
@@ -1,12 +1,75 @@
 @use "@material/select/mdc-select.scss";
-@use "@finastra/fds-theme"as fds;
+@use "@finastra/fds-theme" as fds;
 
 :host {
   display: block;
+  @include fds.margin(4, top);
```
dorsaffrigui, line 6: Do we really need a margin top ? in the data table example the select isn't well positioned in relation to the other elements

Screenshot 2022-08-29 at 11 05 44

bcldvd (approved):

dorsaffrigui (approved):

PR #856, opened 2022-08-30
bugfix#847 fix item bar size

### LINKED ISSUE #847: [Icon bar]: change icon bar item size
**Describe the bug**
fds-icon-bar-item size shoud be 32 * 32, actual size is 38 * 38.

**Screenshots**

Screenshot 2022-08-24 at 11 03 55

Pi5rr5:
Fix #847

### DIFF
--- a/libs/web-components/icon-bar/src/icon-bar.ts
+++ b/libs/web-components/icon-bar/src/icon-bar.ts
@@ -18,16 +18,16 @@ export class IconBar extends LitElement {
   @property({ type: Boolean }) large = false;
 
   @property({ type: Boolean })
-  removeNotification= false;
+  removeNotification = false;
 
-  private footer= false;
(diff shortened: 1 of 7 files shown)


libs/web-components/icon-bar/src/icon-bar.ts
```diff
@@ -18,16 +18,16 @@ export class IconBar extends LitElement {
   @property({ type: Boolean }) large = false;
```
dorsaffrigui, line 18: Could you please change the name "large" to "withLabels" ?

Pi5rr5: Changed for showLabels to keep the logic of show/hide.

dorsaffrigui (approved):

ttalbot (approved):

PR #862, opened 2022-09-13
feature(wc): fds guided tour

code2nguyen:
- [x] Remove menu + start tour current mechanist 
- [x] Add Start Tour function and start delay parameter
- [x] Add Skip tour function
- [x] Skill tour dialog 
- [x] Add some ajustement offset/margin for tour item from JSON data

Inspired ref:
- https://github.com/wwayne/react-tooltip/blob/master/src/index.js
- https://github.com/shipshapecode/shepherd
- https://userguiding.com/blog/introjs-alternatives/


rayah:
- [ ] To fix Text alignment with logo in app bar image
- [ ] To use primary and secondary buttons instead of only primary
image

bcldvd:
Once clicking on the next (or back) button, the following steps have the same button in "focused" state.
Is it possible not to have that behavior ?

bcldvd:
image

When clicking on the next (or back) button, following steps have their button in "focused" state.
Is it possible not to have that behavior ?

bcldvd:
When having finished the tour once, and clicking "Start tour" again, it goes back to the last step.
Shouldn't we want to go back to the first step @rayah  ?
Or is this configurable from the component itself, and the decision of the team using the tour ?

bcldvd (approved):

dorsaffrigui (approved):

PR #883, opened 2022-10-03
feat(textfield): applying v2 outline style

ttalbot:
Only adding text field: label inside, normal and dense sizes.

image

Should not affect v1 "fill" appearance, developers needs to opt-in by specifying `appearance="outline"`

### DIFF
--- a/apps/angular-test-app/src/app/components/ui-elements-demo/ui-elements-demo.component.html
+++ b/apps/angular-test-app/src/app/components/ui-elements-demo/ui-elements-demo.component.html
@@ -668,7 +668,6 @@
       </mat-form-field>
     </div>
   </div>
-
 </div>
 
 <div>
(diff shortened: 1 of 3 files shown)


apps/angular-test-app/src/app/components/ui-elements-demo/ui-elements-demo.component.html
```diff
@@ -494,20 +494,22 @@ <h3 class="uxg-h3">Slide Toggle</h3>
   <h3 class="uxg-display-3">Inputs</h3>
   <h5 class="uxg-h5">Fill Input</h5>
```
rbordeanu, line 495: Shouldn't we keep the fill appearance inside the Fill Input section? ;)

Maybe showcase both fill and outline, in separate sections.

rbordeanu (changes requested):

rbordeanu:
Not related to this PR, but it bothers me a bit that not all of these rows are left aligned :)

Either use `example-row` class on all of them, or on none.

Screenshot 2022-10-06 at 15 11 38

dorsaffrigui (approved):

PR #891, opened 2022-10-07
feature(fds-data-table): checkbox column, chip column, linear progres…

Ping-YUAN:
Enhance data table

1. select: multiple select checkbox, single select radio box.
2. new columns:  link, chip, linear progress, date, type double, cell template.
3. dense version
4. enable style for rows and columns
5. style align to design
6. Doc update

Screenshot 2022-10-24 at 15 12 29


Pi5rr5:
url color need to be fixed 
image

Pi5rr5:
the hover color is not correct, and the hover on the alternated row is not visible
image

Pi5rr5:
Can you add a story with multi-select checkbox and single select radio ?

dorsaffrigui (commented):
Could you add an fds color to the fds-data-table-link class ?
Screenshot 2022-10-24 at 15 46 13

libs/web-components/data-table/src/data-table-styles.scss
```diff
@@ -25,43 +23,199 @@
...
+  .fds-data-table {
+    .mdc-data-table__table {
 
-  .fds-data-table{
-    .mdc-data-table__table{
-      thead,tbody tr{
-        th, td, fds-icon-button{
-          @include fds.text-color(on-background);
+      background: var(--fds-surface, #FFFFFF);
+      border: var(--fds-data-table-border-width, 1px) solid var(--fds-outline, rgba(0, 0, 0, 0.12));
```
ttalbot, line 30:
```suggestion
      @include fds.property('border-color', outline);
      border-width: var(--fds-data-table-border-width, 1px);
      border-style: solid;
```

libs/web-components/data-table/src/data-table-styles.scss
```diff
@@ -25,43 +23,199 @@
...
+
+      .mdc-data-table__header-row,
+      .mdc-data-table__row{
+        height: 56px;
+      }
+
+      thead tr{
+        background-color: var(--fds-surface-hover, rgba(105,78,214, 0.04));
```
ttalbot, line 38:
```suggestion
        @include fds.bg-color(surface-hover);
```

libs/web-components/data-table/src/data-table-styles.scss
```diff
@@ -25,43 +23,199 @@
...
+      }
+      &.center div {
+          display: flex;
+          place-content: center;
+          align-items: center;
+      }
+      }
+
+      tbody tr:hover{
+        background-color: var(--fds-surface-hover, rgba(105,78,214, 0.04));
```
ttalbot, line 58:
```suggestion
        @include fds.bg-color(surface-hover);
```

libs/web-components/data-table/src/data-table-styles.scss
```diff
@@ -25,43 +23,199 @@
...
+      thead,
+      tbody {
+        fds-icon{
+          vertical-align: middle;
+          max-width: 18px;
+          --fds-icon-size: 18px;
+        }
+        // width: calc(100% - 1em);
+        tr:nth-child(even){
+          background-color: var(--fds-surface-hover, rgba(105,78,214, 0.04));
```
ttalbot, line 70:
```suggestion
          @include fds.bg-color(surface-hover);
```

libs/web-components/data-table/src/data-table-styles.scss
```diff
@@ -25,43 +23,199 @@
...
+            .mdc-data-table__cell--radiobox-container {
+              display: flex;
+              place-content: center;
+            }
+          }
+        }
+      }
+
+      .mdc-data-table__row--selected {
+        background-color: var(--fds-surface-selected) !important;
```
ttalbot, line 139: If the theme is not loaded, it won't work because there's no fallback value for the custom property.
Find a way to use mixin here as well.
When you use `!important`, that's the end of the game.
No one will be able to override this style. Try to be more specific than mdc instead, that will allow further customisation later.

ttalbot (dismissed):

libs/web-components/data-table/src/data-table-with-pagination.ts
```diff
@@ -27,76 +27,86 @@ export class DataTableWithPagination extends LitElement{
...
-    override render(){
+    override render() {
         return html`
         <div style="display:flex;
-                    flex-direction: column;
-                    justify-content: center;
-                    place-content: flex-end flex-start;
-                    align-items: flex-end;
-                    ">
+                            flex-direction: column;
```
ttalbot, line 74: Is it necessary to put styling in HTML rendering instead of SCSS file?

Ping-YUAN: it's actually just an example component to demo data with pagination. no more css need in the scss file so just quickly added there/

ttalbot (approved):

dorsaffrigui (approved):

PR #921, opened 2022-11-03
fix(fds-data-table): fix fds-data-table-with-pagination style

Ping-YUAN:


blminami (commented):
On `onFdsDataTableWithPaginationRowSelected` event, this is the object that I received for the row selected .. I think it would be great to remove one level of depth for `detail`, from `detail.data.data[0]` to `detail.data[0]`.

It can be done here: libs/web-components/data-table/src/data-table-with-pagination.ts
` detail: {
         data: this.selected,
}`

to

`detail: this.selected` 

OR

✨✨`data: this.selected.data[0]` if we want to access it like `data[fieldName]` (I would go for this version tbh, it's easier to read)✨✨


image

blminami:
As a general note, it would be great to have the actions / events on storybook as well. Otherwise, the developer has to check the component implementation in order to interact with the component.

If such section exists, can you point it out to me? I was not able to find it :(

ttalbot (approved):

PR #924, opened 2022-11-07
Add support for changing appearance for angular components

achilehero:
Add support for changing appearance for angular components:
- filter panel
- search input

### DIFF
--- a/apps/angular-test-app/src/app/components/filter-panel-demo/filter-panel-demo.component.html
+++ b/apps/angular-test-app/src/app/components/filter-panel-demo/filter-panel-demo.component.html
@@ -36,7 +36,7 @@
 
         <div class="filter-panel-tags">
           <div class="filter-panel-tags-title">
-            <h5 class="uxg-h5">Tags</h5>
+            <h5 class="uxg-h5">Tags, appearance type outline</h5>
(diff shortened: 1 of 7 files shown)


ttalbot (approved):

blminami (commented):
On hover, wouldn't it be better to use primary color (or a variation of primary color)?

image

achilehero:
> On hover, wouldn't it be better to use primary color (or a variation of primary color)?
> 
> image

I tend to agree to that. @ttalbot , do we do that for all inputs? Ue primary color for the border on hover? Just like we have it for the focused input?

achilehero:
So we should leave the border black for the outlined input on hover, then?

achilehero:
OK. So, @blminami , you have the answer. We have to leave it like the original mat input.

blminami (approved):

rbordeanu (approved):

achilehero:
Do we need @adrian-apetrei 's review in order to merge this? 😃

adrian-apetrei:
> Do we need @adrian-apetrei 's review in order to merge this? 😃

Looks good to me, keep up the good work! 😆

achilehero:
Thanks, Adi! 😄

PR #944, opened 2022-11-10
fix(fds-data-table): fix header when scroll and fix table column width.

Ping-YUAN:


libs/web-components/data-table/src/data-table-styles.scss
```diff
@@ -24,34 +24,40 @@
...
-      &.center div {
-          display: flex;
-          place-content: center;
-          align-items: center;
-      }
+      thead {
+        position: sticky;
+        top: 0px;
+        z-index: 1;
+        tr{
```
blminami, line 42: format code, there should be an empty space between the selector and the bracket `tr {`

blminami (approved):

ttalbot (approved):

PR #998, opened 2022-12-07
fix(fds-button): disabled button triggering click events

### LINKED ISSUE #993: [web-component]:Button still works even if disabled
**Describe the bug**
Button still work even if disabled

**Code sample**
`<fds-button type="submit" label="submit" (click)="submit()" [attr.disabled]="packageForm.invalid? true: null"></fds-button>`

**Expected behavior**
Button not work if disabled

**Screenshots**
image

dorsaffrigui:
Fix #993


ttalbot (approved):
Yes, but a lot of styling change for just 1 line of bug fix. We need to fix that.

libs/web-components/button/src/base-styles.scss
```diff
@@ -18,7 +18,7 @@
   @include fds.padding-x(3);
   @include fds.padding-y(2);
 
-  --fds-icon-color: var(--fds-on-primary, #694ED6);
+  --fds-icon-color: var(--fds-on-primary, #694ed6);
```
ttalbot, line 21:
```suggestion
  --fds-icon-color: var(--fds-on-primary, #694ED6);
```

Pi5rr5 (approved):

PR #1010, opened 2022-12-13
fix(toggle-group): update selection and hover styles

ttalbot:
update selection size and position when fonts are loaded.
use 'selected-index' attribute name
Add hover and focus states

### DIFF
--- a/libs/web-components/button-toggle-group/src/button-toggle/styles.scss
+++ b/libs/web-components/button-toggle-group/src/button-toggle/styles.scss
@@ -6,13 +6,19 @@
   @include fds.padding-x(3);
 
   display: inline-flex;
-  align-items: baseline;
+  align-items: center;
   border: none;
+  box-sizing: border-box;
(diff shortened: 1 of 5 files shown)


libs/web-components/button-toggle-group/src/button-toggle/styles.scss
```diff
@@ -6,13 +6,19 @@
...
   outline: none;
   background: transparent;
   cursor: pointer;
   transition: color 0.2s ease;
   user-select: none;
+  height: 36px;
+  border-radius: 16px;
+  transition-property: background-color;
+  transition-duration: 0.3s;
+  background-color: transparent;
```
ttalbot, line 21:
```suggestion
  height: 36px;
  border-radius: 16px;
  background-color: transparent;
  transition-property: background-color;
  transition-duration: 0.3s;
```

bcldvd (approved):

Issue #1060, opened 2023-02-08
[chips] padding are too large when using leading icon


ttalbot:
**Describe the bug**
chips padding are too large with leading icon

**To Reproduce**

1. Go to ['...'](https://finastra.github.io/finastra-design-system/?path=/story/data-display-chip--icon&args=icon:public)

**Expected behavior**
padding between chip border and icon should be 6px

**Screenshots**

Actual:
![image](https://user-images.githubusercontent.com/371041/217575466-2fc16fd5-5b84-4b0a-ae72-2265b742d556.png)
Design:
![image](https://user-images.githubusercontent.com/371041/217575621-b73094f5-c980-4264-9550-63cd96e705d8.png)

Issue #931, opened 2022-11-07
[chips] icon in dense size is too large


ttalbot:
**Describe the bug**
When using dense chips with left icon, it almost touch the borders, which does not look good

**To Reproduce**

1. Go to ['Storybook'](https://finastra.github.io/finastra-design-system/?path=/story/data-display-chip--dense)
2. See error

**Expected behavior**
Icon should be smaller, around 12px instead of 14px.

**Screenshots**
![image](https://user-images.githubusercontent.com/371041/200354789-26131806-e30a-451e-85b1-8315be4a1881.png)

**Additional context**
Error is also coming from Figma design, needs a fix here as well.

PR #1063, opened 2023-02-09
fix spacings for chips

achilehero:
Fixes for #931 and #1060

### DIFF
--- a/libs/web-components/chip/src/styles.scss
+++ b/libs/web-components/chip/src/styles.scss
@@ -75,7 +75,7 @@
   }
 }
 
-:host([icon]) {
+:host([icon]:not([icon=''])) {
   .mdc-button {
     padding-left: 8px;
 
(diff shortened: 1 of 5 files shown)


libs/web-components/chip/src/styles.scss
```diff
@@ -11,6 +11,7 @@
 
   .mdc-button {
     height: 24px;
+    @include fds.padding-x(2);
```
ttalbot, line 14: Use 12px instead of 8px. There's no token for 12px 🤷

libs/web-components/chip/src/styles.scss
```diff
@@ -76,16 +77,18 @@
 :host([dense]) {
   .mdc-button {
     height: 18px;
+    padding: 0 6px;
```
ttalbot, line 80: 6px is only between border and Icon, you should keep 12px when there's no icon (between border and label)

ttalbot (changes requested):
To summarize:

normal size chip:
border <-12px-> label <-12px-> border
border <-8px-> icon <-8px-> label <-12px-> border

dense size chip:
border <-12px-> label <-12px-> border
border <-6px-> icon <-6px-> label <-12px-> border

ttalbot (approved):

PR #1117, opened 2023-08-07
allow to use custom logo

BeataKr:


BeataKr:
no need, user can just set own --fds-logo

